PR #24, opened 2018-07-06
Small text improvements + adding copy buttons to text blocks

choldgraf:
This makes small edits to some text as I was going through the instructions (primarily the gcloud ones)

It also implements some simple javascript/CSS that creates copy buttons to the right of each code block. For an example of what this looks like in action, see:

http://predictablynoisy.com/the-littlest-jupyterhub/


yuvipanda:
w00t, tyvm @choldgraf :)

PR #25, opened 2018-07-06
adding google instructions and faq

choldgraf:
A short add-on to #24 since you beat me to the merge :-)

### DIFF
--- a/docs/troubleshooting/faq.rst
+++ b/docs/troubleshooting/faq.rst
@@ -1,0 +1,16 @@
+Frequently Asked Questions
+==========================
+
+I stopped and started my VM, and now JupyterHub doesn't work!
+----------------------------------------------------------------------
+
+Depending on the VM provider that you're using, stopping your VM might
+cause a new External IP address to be generated. If this happens, then your
+old IP address will no longer work.
+
+Figure out what the new External IP address is, then SSH into it and
+restart your JupyterHub with::
+
+    sudo systemctl restart jupyterhub
+
+And the new External IP should now start working.
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -57,7 +57,8 @@ guides help you find what is broken & hopefully fix it.
    :titlesonly:
 
    troubleshooting/logs
-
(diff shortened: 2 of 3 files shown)


docs/troubleshooting/faq.rst
```diff
@@ -0,0 +1,16 @@
...
+==========================
+
+I stopped and started my VM, and now JupyterHub doesn't work!
+----------------------------------------------------------------------
+
+Depending on the VM provider that you're using, stopping your VM might
+cause a new External IP address to be generated. If this happens, then your
+old IP address will no longer work.
+
+Figure out what the new External IP address is, then SSH into it and
```
yuvipanda: You don't need to do this.

docs/troubleshooting/faq.rst
```diff
@@ -0,0 +1,16 @@
+Frequently Asked Questions
+==========================
+
+I stopped and started my VM, and now JupyterHub doesn't work!
```
yuvipanda: I think we should also be a lot more specific here, since there is a lot of reasons this could be the case!

Maybe phrase it something else than "Doesn't work" and focus it on Google Cloud?

docs/troubleshooting/faq.rst
choldgraf: hmmm, it's the only way I could get it working in my case :-)

docs/troubleshooting/faq.rst
choldgraf: yeah, perhaps this section should be in the google cloud part of the guide in general, instead of in faq?

yuvipanda: yeah. But not the tutorial though, I think the tutorial should all be the same structure. Maybe start a provider specific tips section of the guides?

docs/troubleshooting/faq.rst
yuvipanda: Interesting. I wonder if this means the service doesn't start automatically on restart?  Can you file a bug separately for that?

PR #34, opened 2018-07-13
remove —no-cache-dir arg

minrk:
caches are good! disabling cache should ~only be used when building docker images

### DIFF
--- a/bootstrap/bootstrap.py
+++ b/bootstrap/bootstrap.py
@@ -87,7 +87,7 @@ def pip_install(prefix, packages, editable=False):
     Set editable=True to add '--editable' to the pip install commandline.
     Very useful when doing active development
     """
-    flags = ['--no-cache-dir', '--upgrade']
+    flags = ['--upgrade']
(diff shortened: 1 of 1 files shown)


yuvipanda:
It puts things in /root, which fails in the init-script stage in some places some of the time. This is why I have it disabled in the bootstrap / installer stage! I'll document that..

yuvipanda:
It looks like conda might still be putting stuff in /root, I'll investigate.

minrk:
Ah, that's weird. Probably down to where $HOME is, which is typicall /root for root. Where does that fail?

yuvipanda:
I had it fail with a 'root file system readonly' error on DigitalOcean with a traceback to /root, but can't reproduce it anymore. I'll just merge this now and see if it pops up again.

yuvipanda:
There's a bunch of '  Failed building wheel for pyyaml' in there now, because wheel building fails. I'm tempted to revert this for now.

Issue #36, opened 2018-07-14
Support additional authenticators


yuvipanda:
Currently we only support dummy & firstuse authenticator.

I want us to support arbitrary authenticators without having to add custom code for each of them. z2jh requires we do a bunch of extrenous work mapping YAML config to authenticator config for it to work. I want us to not have to do that, somehow.

minrk:
We could rely on standard config.py so users could follow the given authenticator's installation and config documentation and use traitlets `load_subconfig` to load file(s) in a particular location:

```python
import os
import glob

for f in glob.glob('/etc/jupyterhub/*_config.py'):
    load_subconfig(f)
```

(or it could be just one supported config file at /etc/jupyterhub/jupyterhub_config.py)

Then the simple case is that users edit `/etc/jupyterhub/jupyterhub_config.py` and it's loaded as additional config by /srv/tlj/jupyterhub_config.py.

yuvipanda:
Fixed by a combination of #37, #78

PR #37, opened 2018-07-15
Dynamically set authenticator properties from YAML

yuvipanda:
We don't wanna explicitly map keys from the YAML to traitlet
config for the authentication - that's a lot of busywork for no
gain. We instead switch to using snake_case everywhere, and
dynamically set traitlet config from YAML config!

Refer #36


yuvipanda:
I hadn't intended to set up codecov to shame me if code coverage dropped, but it did and i added more tests! I call this a win!

yuvipanda:
I'm also going to say that we don't merge new PRs until they're documented, so I'm going to add docs to this before we can merge it.

PR #51, opened 2018-07-17
[wip] enable sudo-less install for admins

minrk:
Goal: members of jupyterhub-admins should be able to install and upgrade packages without sudo.

The main change is using `setfacl` and `setgid` on the top-level directory to ensure that new files and directories in the env are owned by `jupyterhub-admins` and group-writable.

Changes:

- merge hub and user env in `/opt/tljh/env`
- move state out of env to `/opt/tljh/state`
- integration tests for permissions:
  - everyone can read every file in the env
  - admins can write any directory
  - neither admins nor users can read or write to state
  - admins can pip install/upgrade
  - users cannot pip install/upgrade

Somewhat unrelated fixes along the way:

- allow conda to be upgraded (fix == version check instead of >= check)

Some weirdnesses:

Not all *files* are group-writable. This is because of the behavior of pip and conda installers, which can make files read-only and/or not group-owned. This is okay for our purposes, because installation/upgrade should never involve *modification* of files, only removal of files and writing of new files. That means that only write-permission on directories is required. Sudo remains available if individual files want to be modified, which should be a rare, advanced use case.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,8 +39,7 @@ jobs:
     integration-test:
         machine: true
         environment:
-            cmd: "python3 .circleci/integration-test.py"
-            PATH: "$HOME/conda/bin:$PATH"
+            cmd: "/opt/conda/bin/python3 .circleci/integration-test.py"
         steps:
             # Download and cache dependencies
             - restore_cache:
@@ -51,13 +50,13 @@ jobs:
                 name: setup python3.6
                 command: |
                     if [[ ! -d "$HOME/conda" ]]; then
-                        curl -L https://repo.continuum.io/miniconda/Miniconda3-4.5.4-Linux-x86_64.sh > miniconda.sh
-                        bash miniconda.sh -b -p $HOME/conda
+                        curl -L https://repo.continuum.io/miniconda/Miniconda3-4.5.4-Linux-x86_64.sh > /tmp/miniconda.sh
+                        sudo bash /tmp/miniconda.sh -b -p /opt/conda
                     fi
 
             - save_cache:
                 paths:
-                    - $HOME/conda
+                    - /opt/conda
                 key: v1-integration-miniconda3-4.5.4
 
             - checkout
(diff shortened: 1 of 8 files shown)


yuvipanda:
Thanks for working on this, @minrk! <3

I'd like to keep the user and hub env separate, primarily as a protection against users accidentally screwing up the hub environment. Especially since the primary way we want users to interact with the hub is via the hub itself, this could be bad.

I'll take a more detailed look tomorrow!

.circleci/config.yml
```diff
@@ -37,49 +37,61 @@ jobs:
                     codecov
 
     integration-test:
-        docker:
-            - image: docker:18.05.0-ce-git
-
+        machine: true
```
yuvipanda: is there a reason this is using machine instead of container?

yuvipanda: ah, I see you're just hacking around it to find out why things are failing :)

yuvipanda:
I'm also now less sure that the amount of complexity (around ACLs and sticky bits) this adds is worth the effort at this stage of the project. How do you feel about delaying this until we get a release out later the coming week?

.circleci/config.yml
minrk: Yup. While the acls work on a 'real' machine, the filesystem on CI has some of the same acl issues that images have. After spending a bit of time, I think we can't reasonably test on Circle if we want to see how a real filesystem would be have.

minrk:
> I'd like to keep the user and hub env separate

Sure! I thought we discussed this yesterday and concluded that the envs should not be separate, but that's easy to revert.

After fighting with Circle for a bit and dealing with weird issues, I think I'll break this up into one or two smaller PRs that are mostly the fixes-along-the-way instead of the original motivating feature.

> How do you feel about delaying this until we get a release out later the coming week?

Makes perfect sense. I think one or two of the smaller fixes should get in, but I think it's a good idea to save the "admin install without sudo" feature for later.

I really wish people cared about filesystems in containers :(.

minrk:
No need to keep this open, since it's pretty hairy to test (easier to do in practice, but CircleCI doesn't give us a realistic environment in which to test).

PR #52, opened 2018-07-17
Fix link to custom tutorial

parente:
Missing an s


yuvipanda:
<3 thank you very much, @parente!

Issue #49, opened 2018-07-17
Add support to install libraries during TLJH installation


geoffbacon:
Currently, to configure the user environment with libraries, we log in as admin and install them one by one. Given that this is likely to be a common procedure, an alternative is to allow for a requirements.txt/environment.yml file to be included in the TLJH installation process.

yuvipanda:
Thank you for reporting this, @geoffbacon!

Partially fixed by #53, which adds requirements.txt support. Would be nice to add environment.yaml support too!

yuvipanda:
We can add environment.yaml support later when someone asks for it!

PR #53, opened 2018-07-18
Add option to install requirements.txt file on install

yuvipanda:
Also add docs about --admin option

Refer #49


yuvipanda:
@geoffbacon what do you think of this?

geoffbacon:
Yep, perfect! 👍

Issue #62, opened 2018-07-19
Supporting ARM architectures


gedankenstuecke:
It would be 💯 to get TLJH running on Raspberry Pis and other small boards. Unfortunately `miniconda` doesn't support their ARM processor architectures, so some changes to the installer are needed. 

As discussed with @yuvipanda a fix for this would be using `virtualenv` instead of `conda`. The user environment should be configurable to use either `conda` or `virtualenv`. Furthermore nodesource also supports ARM architectures.

yuvipanda:
Thank you for testing and reporting this early, @gedankenstuecke!

Our plans now are:

1. Require Ubuntu 18.04 as minimum, since system python here is 3.6
2. Use nodesource + venv for the hub environment unconditionally.
3. Use conda as default user environment in everything except ARM, where venv is user environment. This is not going to be user toggleable - we'll switch based on architecture.
4. Integration test ARM with qemu. This is gonna be tricky but absolutely necessary. We probably won't run unit tests in it.
5. Use *only* pip for installing packages into the user environment / hub environment in the default install. This reduces split between ARM & x86 environments.

I think this covers it.

yuvipanda:
Awesome, @gedankenstuecke! Can you write up a small script similar to https://github.com/jupyterhub/the-littlest-jupyterhub/blob/master/.circleci/integration-test.py that helps run TLJH inside a qemu / arm container?

gedankenstuecke:
Yeah, I tried adapting the whole thing here with a custom Dockerfile and integration-test.py: https://github.com/gedankenstuecke/the-littlest-jupyterhub/commit/de1c7ce2a3df5aa53c0120f645a8eb90c95c0d1a

The building of the image seems to work out fine, but then when trying to start the container it dies right away and I couldn't yet figure out what's going wrong here.

yuvipanda:
Things that need to happen here:

1. Allow switching between conda & venv for user environment at install time
2. Add tests for both 
3. Add tests for the littlest jupyterhub in debian stretch images
4. Add tests (with QEMU) on ARM architectures

scparker:
Sounds like a good JupyterCon paper!  I'll check it out later today...

pisymbol:
Is this dead?

yuvipanda:
@pisymbol nobody is currently working on it, unfortunately :( The core set of tasks needed haven't changed though.

cdibble:
[...]

PR #64, opened 2018-07-20
Use venv for base hub environment

yuvipanda:
- TLJH should support raspberry pi, which runs ARM. conda does
  not support ARM.
- Get nodejs from nodesource instead of conda or default repositories.
  Default repositories get out of date pretty quickly.
- Install CHP from npm

Refer #62

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,8 +19,8 @@ jobs:
             # Download and cache dependencies
             - restore_cache:
                 keys:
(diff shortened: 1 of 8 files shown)


yuvipanda:
The conda unit tests are failing, since the environment has changed a lot. I'm going to redo the unit tests to run in an environment more similar to the final production - running in a venv themselves, but calling out to conda from the user environment.

psychemedia:
64 bit Raspberry Pi OS should be able to install [miniforge](https://github.com/conda-forge/miniforge)?

Issue #57, opened 2018-07-19
update documentation on disk size


norcalbiostat:
Step 13 says: 
> In Google Cloud, the higher your disk size the faster your disk is. The default (10GB) is pretty low, so you might want to increase it.

Please add guidance or suggestions on how much to increase it based on # of students or hours.

yuvipanda:
Closed by #68

Issue #56, opened 2018-07-18
update documentation on guidance for choosing a VM size


norcalbiostat:
Add more specific info and guidelines to step 10 on https://the-littlest-jupyterhub.readthedocs.io/en/latest/tutorials/google.html 
The link to the google cloud docs is not helpful. 
Suggestions on what would be best for 30 students, 50, 75 etc. 
Do we need more than 1 CPU to handle multiple students login? 
How many hours are typical? Instances are running 24 hours a day right? 
This type of info is helpful for the new user.

yuvipanda:
Closed by #68. Let us know if that's not good enough, @norcalbiostat

Issue #48, opened 2018-07-17
Ballpark estimates for VM resources in the docs


geoffbacon:
It'd be great to have suggestions in the docs on the number, size and type of resources for typical use cases. For example, "if you expect 50 users crunching big numbers, you might need at least X CPUs with Y RAM, but if you're only having 5 users learning to code for the first time, you probably only need 1 CPU with this much RAM". I can imagine a lot of users (including myself) don't have enough background to get this right.

yuvipanda:
Closed by #68

PR #68, opened 2018-07-22
Add a how-to guide on selecting VM Memory / CPU / Disk size

yuvipanda:
Ref #57, #56, and #48

### DIFF
--- a/docs/howto/resource-estimation.rst
+++ b/docs/howto/resource-estimation.rst
@@ -1,0 +1,77 @@
+.. _howto/resource-estimation:
+
+===================================
+Estimate Memory / CPU / Disk needed
+===================================
+
+This page helps you estimate how much Memory / CPU / Disk the server you install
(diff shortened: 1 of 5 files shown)


yuvipanda:
@geoffbacon @norcalbiostat There's now more docs on this at https://the-littlest-jupyterhub.readthedocs.io/en/latest/howto/resource-estimation.html. Would love feedback on how it reads :)

geoffbacon:
Yeah, this is really useful! One minor suggestion I have is to provide a hint on how to estimate memory usage for a single user. Even just linking to something like [this](https://timothymonteath.com/monitoring-memory-usage-in-jupyter-notebooks/) .

yuvipanda:
@geoffbacon awesome! I think I'll add nbresuse instead...

geoffbacon:
:+1:

norcalbiostat:
Looks great, thank you!

PR #87, opened 2018-07-31
Load arbitrary .py config files from a conf.d dir

yuvipanda:
We want to keep the amount of custom code we support extremely
small. This provides a nice escape hatch for everything else.

- [x] Add documentation on how to use this
- [x] Add a test for this

### DIFF
--- a/tljh/jupyterhub_config.py
+++ b/tljh/jupyterhub_config.py
@@ -3,8 +3,8 @@ JupyterHub config for the littlest jupyterhub.
 """
 import copy
 import os
-
 import yaml
+from glob import glob
 
 from systemdspawner import SystemdSpawner
 from tljh import user, configurer
@@ -47,3 +47,9 @@ if os.path.exists(config_overrides_path):
 else:
     config_overrides = {}
 configurer.apply_config(config_overrides, c)
+
+# Load arbitrary .py config files if they exist.
+# This is our escape hatch
+extra_configs = sorted(glob(os.path.join(INSTALL_PREFIX, 'config.d', '*.py')))
+for ec in extra_configs:
+    load_subconfig(ec)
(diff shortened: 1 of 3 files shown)


tljh/jupyterhub_config.py
```diff
@@ -49,3 +50,9 @@ def start(self):
 else:
     config_overrides = {}
 configurer.apply_config(config_overrides, c)
+
+# Load arbitrary .py config files if they exist.
+# This is our escape hatch
+extra_configs = sorted(glob(os.path.join(INSTALL_PREFIX, 'config.d', '*.py')))
```
minrk: Let's call it 'jupyterhub_config.d' to make it clear what it configures

yuvipanda: That makes sense, @minrk! Will do.

consideRatio: Oh - does `.d` indicate for historial reasons that a file is a partial file?

yuvipanda: @consideRatio yep! I picked it up from linux distros like debian - you have a /etc/sudoers file and a /etc/sudoers.d directory that can contain partials.

consideRatio: Ah! I love when a piece like that falls into place! :heart: @minrk @yuvipanda

yuvipanda: Done, @minrk

minrk:
Should we create the `jupyterhub_config.d` directory with a README during install?

yuvipanda:
@minrk since i consider this an 'advanced use case', I think it's fine to not do that. But I'll add a note to the docs about needing to create the directory.

yuvipanda:
No tests for this, need to land #97 first.

PR #88, opened 2018-07-31
Fix traefik config reload

yuvipanda:
Simply reloading traefik is not enough - we need to actually
write the new traefik.toml file.

### DIFF
--- a/tljh/config.py
+++ b/tljh/config.py
@@ -15,7 +15,7 @@ import sys
 import argparse
 from ruamel.yaml import YAML
 from copy import deepcopy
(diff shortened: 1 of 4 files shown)


yuvipanda:
@minrk I'm not entirely sure what the right thing to do here is. We do need this for HTTPS to work though, I think.

minrk:
@yuvipanda you're right! I fixed it by making the path constants importable from `tljh.config`, so it's called with STATE_DIR

yuvipanda:
Awesome, thanks @minrk

PR #89, opened 2018-07-31
tests and fixes in tljh-config

minrk:
- add remove-item to cli
- fix setting non-string values (int, bool, float)
- show help when no action is given
- test coverage for cli

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -1,11 +1,30 @@
 """
 Test configuration commandline tools
 """
-from tljh import config
-from contextlib import redirect_stdout
-import io
-import pytest
+
+from importlib import reload
+import os
 import tempfile
+from unittest import mock
+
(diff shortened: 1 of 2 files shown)


yuvipanda:
Awesome, thanks @minrk!!!

chmendoza:
Hi there!

I am trying to use this to delete a user from cli:

```
sudo tljh-config remove-item users.admin mendoza
sudo tljh-config reload
```

after I do that, this is the content of `/opt/tljh/config/config.yaml`:

```
users:
  admin: []

```

However, I can still log as mendoza. In addition, there are other users that I created through the web interface that are not showing up on the config.yaml file. See this screenshot (after executing the above commands):

image

BTW, I am deleting the user to then creating it again, as a workaround to reset the password :)

Issue #54, opened 2018-07-18
Add detailed tutorial on setting up Google Authenticator


yuvipanda:
Should be completely step by step, with images and links to people for people who have no prior experience

PR #91, opened 2018-07-31
small updates to the docs

choldgraf:
This is a general sweep of the docs at @yuvipanda 's request :-)

I made small edits, rearrangements, etc as I went through. Feel free to add comments, clarifications, or questions and I can address in this PR. Alternatively I'm happy to spot-check things in subsequent PRs.

In general, these docs are really, really nice. I think we should consider a Z2JH refactor with this as an inspiration.


docs/howto/auth/ldap.rst
```diff
@@ -0,0 +1,46 @@
+.. _howto/auth/ldap:
```
yuvipanda: I think we should either make this be fully complete and talk about at least the required parameters of LDAPAuthenticator, or remove it.

docs/howto/resource-estimation.rst
```diff
@@ -40,9 +40,16 @@ Maximum memory allowed per user
 
 Depending on what kinda work your users are doing, they will use different amounts
 of memory. The easiest way to determine this is to run through a typical user
-workflow yourself, and measure how much memory is used. Add 20-40% headroom for
-users to 'play around', and that should be the maximum memory allowed per user.
-The system will prevent users from using more memory than this.
+workflow yourself, and measure how much memory is used.
+
+For example, you can begin running a Jupyter Notebook session on your JupyterHub, then open a
```
yuvipanda: This is great for now! I wanna add nbresuse to the default and use that instead though.

docs/index.rst
```diff
@@ -3,13 +3,13 @@ The Littlest JupyterHub
 =======================
 
 A simple `JupyterHub <https://github.com/jupyterhub/jupyterhub>`_ distribution for
-a small (0-50) number of users on a single server.
+a small (0-100) number of users on a single server.
```
yuvipanda: YES! Have to decide what's the number we should use. 100 seems ok for now.

docs/topic/authenticator-configuration.rst
```diff
@@ -24,6 +24,12 @@ example.
 Configuring the authenticator
 =============================
 
+Some authenticators have unique configuration options. This section covers a
```
yuvipanda: This document talks about configuring generic authenticators without making specific comments about anything - LDAPAuthenticator is only used as an example. If you found that confusing, we should restructure this quite a bit - however, I think the characterization as 'a few common ones' is not correct.

For auth, I think we will have three types of documentation:

1. Tutorials with step by step for an authenticator, including screenshots (Such as #54). We don't have any yet.
2. How-to guides for specific authenticators, such as what we have for dummyauthenticator
3. This document, which is the 'fall through' when we don't have tutorials or howtos for an authenticator.

I think LDAPAuthenticator is common enough to be a HOWTO. Maybe we should use something else as an example here?

We should also probably put this comment in contributing/doc.rst.

docs/topic/security.rst
```diff
@@ -38,16 +38,22 @@ command on the terminal. No password required.
...
 ===========================
 
 When you delete users from the JupyterHub admin console, their unix user accounts
 are **not** removed. This means they might continue to have access to the server
 even after you remove them from JupyterHub. Admins should manually remove the user
-from the server & archive their home directories as needed. If the user removed
+from the server & archive their home directories as needed. For example, the
+following command deletes the user ``yuvi``.
+
+.. code-block::
```
yuvipanda: I think this should be:

sudo userdel jupyter-yuvipanda

to delete a user named 'yuvipanda' from the jupyterhub.

docs/howto/auth/ldap.rst
choldgraf: the main reason I added it there was because I was confused that LDAP was getting "special treatment" on the main authenticator page. We could just add contextualizing language there saying "we're just calling out LDAP for example" or something. WDYT?

docs/howto/resource-estimation.rst
choldgraf: yeah I was gonna suggest the same thing, +1 from me on adding by default

docs/topic/authenticator-configuration.rst
choldgraf: see my comments above re: restructuring. You are correct that the main reason I spun it out into its own section was because I found it confusing that LDAP was getting a lot of extra attention. I think if we just made it more clear that LDAP was just for example, it could be better.

It sounds like you have a clearer view of how the pieces fit together here, do you wanna make a push to this PR with how you'd like it to be structured?

docs/topic/security.rst
choldgraf: ooops! good point

yuvipanda:
@choldgraf I've made all the changes I wanted to, lmk if this is ok with you!

choldgraf:
woot! I am +1 on merge!

docs/index.rst
consideRatio: Wow thats awesome!!! :D

PR #94, opened 2018-07-31
Symlink tljh-config to /usr/local/bin

yuvipanda:
This lets `sudo -E` work when calling tljh-config. It currently
does not, since we do not add the hub venv to PATH

 - [x] Add tests

### DIFF
--- a/integration-tests/test_hub.py
+++ b/integration-tests/test_hub.py
@@ -10,6 +10,10 @@ import grp
 import sys
 
 
+# Use sudo to invoke it, since this is how users invoke it.
+# This catches issues with PATH
+TLJH_CONFIG_PATH = ['sudo', '-E', 'tljh-config']
+
 def test_hub_up():
     r = requests.get('http://127.0.0.1')
     r.raise_for_status()
@@ -45,10 +49,9 @@ async def test_user_admin_add():
     hub_url = 'http://localhost'
     username = secrets.token_hex(8)
 
-    tljh_config_path = [sys.executable, '-m', 'tljh.config']
 
-    assert 0 == await (await asyncio.create_subprocess_exec(*tljh_config_path, 'add-item', 'users.admin', username)).wait()
-    assert 0 == await (await asyncio.create_subprocess_exec(*tljh_config_path, 'reload')).wait()
+    assert 0 == await (await asyncio.create_subprocess_exec(*TLJH_CONFIG_PATH, 'add-item', 'users.admin', username)).wait()
+    assert 0 == await (await asyncio.create_subprocess_exec(*TLJH_CONFIG_PATH, 'reload')).wait()
 
     # FIXME: wait for reload to finish & hub to come up
     # Should be part of tljh-config reload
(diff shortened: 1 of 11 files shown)


yuvipanda:
Trying to figure out how to document it. I'm thinking of a 'what does installing TLJH do to my system?' document.

integration-tests/test_hub.py
```diff
@@ -45,10 +49,9 @@ def test_hub_up():
     hub_url = 'http://localhost'
     username = secrets.token_hex(8)
 
-    tljh_config_path = [sys.executable, '-m', 'tljh.config']
 
-    assert 0 == await (await asyncio.create_subprocess_exec(*tljh_config_path, 'add-item', 'users.admin', username)).wait()
```
minrk: might be useful to define:

```python
async def check_call(command):
    """Call a subprocess and wait for it asynchronously

    Returns the exit code of the process
    """
    process = await asyncio.create_subprocess_exec(*command)
    return (await process.wait())

def tljh_config(*args):
    """Call tljh-config with arguments

    returns the exitcode of the process
    """
    return check_call(TLJH_CONFIG_PATH + list(args))

```

to be re-used, to avoid having to re-type the `await (await...` lines so often

minrk: This code could also be simplified by calling the blocking `subprocess.check_call`, since there's no benefit to being async here, as far as I can tell.

tljh/installer.py
```diff
@@ -275,6 +275,24 @@ def ensure_jupyterhub_running(times=4):
...
+
+    We symlink to /usr/local/bin to 'fix' this. /usr/local/bin is the appropriate
+    place, and works with sudo -E
+    """
+    tljh_config_src = os.path.join(prefix, 'bin', 'tljh-config')
+    tljh_config_dest = '/usr/local/bin/tljh-config'
+    if not os.path.exists(tljh_config_dest):
+        # If this exists, we leave it alone. Do *not* remove it,
+        # since we are running as root and it could be anything!
+        os.symlink(tljh_config_src, tljh_config_dest)
```
minrk: If the file exists, probably worth checking the link and warning about it?

Possible cases to test:

- symlink exists and already points to our tljh-config
- symlink exists and points to another tljh install that exists (may not work due to version mismatch)
- symlink exists but target doesn't (definitely won't work)

yuvipanda: So:

1. Symlink exists and points to our tljh-config - do nothing.
2. *something* else exists - warn, but don't abort?
3. symlink exists but target doesn't - force our own symlink?

integration-tests/test_hub.py
yuvipanda: Is it ok if I punt this to a future PR? I'd like to keep using the async variants, since these tests need to be async anyway (for the hubtraf calls)

tljh/installer.py
minrk: Yeah, I think that's sensible

yuvipanda: @minrk I think I implemented a variant of it now.

integration-tests/test_hub.py
minrk: Yes, absolutely! I just noticed a lot of duplicate code that could be simplified. async subprocess calls are very rarely beneficial and a lot more verbose.

Issue #92, opened 2018-07-31
Add integration test for HTTPS


yuvipanda:
While setting up let's encrypt based integration testing is probably too hard, we can use manual TLS for integration testing.

1. Generate a certificate pair
2. Set up TLJH to use it
3. Verify that we have it

PR #96, opened 2018-08-01
unittests for traefik

minrk:
This isn't the integration tests requested in #92, but it gets coverage for traefik.py. In particular, this validates that we are creating valid toml, but in no way validates that it tells traefik to do what we want it to.

 - [x] Add tests

includes #95


yuvipanda:
I like it!

yuvipanda:
oooh, more emojis! :ship:

PR #107, opened 2018-08-03
[MRG] Adding CI for documentation and fixing docs warnings

betatim:
 - [ ] Add tests

This adds a CircleCI job that builds the documentation to make sure it builds. It also turns all warnings into errors. The changes to the docs are there to fix the warnings that are now errors.

### DIFF
--- a/docs/contributing/docs.rst
+++ b/docs/contributing/docs.rst
@@ -60,8 +60,8 @@ Then from the ``docs`` directory, build the HTML:
 
      $ make html
 
-To get started contributing, you'll want to read the `reStructuredText
-reference <http://www.sphinx-doc.org/en/master/usage/restructuredtext/>`_.
+To get started contributing, you'll want to read the :ref:`reStructuredText
+reference <sphinx:rst-index>`
 
 Your locally-built documentation will be themed differently than the
 documentation at `the-littlest-jupyterhub.readthedocs.io <https://the-littlest-jupyterhub.readthedocs.io>`_.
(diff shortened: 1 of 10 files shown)


docs/contributing/docs.rst
```diff
@@ -50,18 +50,18 @@ __ http://docutils.sourceforge.net/
...
 
 Then from the ``docs`` directory, build the HTML:
 
-.. console::
+.. code-block:: console
 
      $ make html
 
-To get started contributing, you'll want to read the :ref:`reStructuredText
-reference <sphinx:rst-index>`.
```
betatim: What kind of reference was this meant to be? Something that goes straight to the sphinx documentation?

docs/index.rst
```diff
@@ -30,7 +30,7 @@ Ubuntu 18.04. We have a bunch of tutorials to get you started.
      tutorials/jetstream
      tutorials/google
 
-- :ref:`tutorials/custom`.
+- :ref:`custom <tutorials/custom>`.
```
betatim: This still leads to an error:
```
/root/repo/docs/index.rst:33:undefined label: tutorials/custom (if the link has no caption the label must precede a section header)
```

I don't understand why :(

docs/contributing/docs.rst
yuvipanda: Yes, it is! I'm not sure why this isn't working. I copied this from the djago docs.

betatim: Maybe it requires some kind of setup in `conf.py`? Can you link me to the source of the django docs, then I could investigate.

betatim: Fixed.

betatim:
Pending circle this is ready to be merged. Probably best to merge before other PRs to avoid having to catch up with new docs breakage that is introduced.

yuvipanda:
<3 thanks, @betatim

PR #111, opened 2018-08-03
Install nbresuse by default

yuvipanda:
Extremely useful to see how much RAM you are using right now

 - [x] Add tests


yuvipanda:
For tests, we should probably validate output of 'jupyter serverextension list --sys-prefix' and 'jupyter nbestension list --sys-prefix'

yuvipanda:
Tests <3

Now for documentation <3.

PR #123, opened 2018-08-10
[MRG] updating content from zexuan's user test

choldgraf:

This PR:

* updates several sections, clarifies points, adds some new content, and rearranges the index page so that it is broken down by topic instead of one flat "howto" list.
* Removes the tutorials section for now (with the idea that we may add it back later when there is more content to fill in that section).
* Adds a CSS rule so the title fits
* Updates some REFs that were broken
* Splits up the docs requirements files into two files so that we can get the docs tests to pass


docs/howto/auth/github.rst
```diff
@@ -9,52 +9,81 @@ GitHub user ID / password. To do so, you'll first need to register an
...
+     :alt: Your client ID and secret
+
+.. important::
+
+   If you are using a virtual machine from a cloud provider and
+   **stop the VM**, then when you re-start the VM, the provider will likely assign a **new public
+   IP address** to it. In this case, **you must update your GitHub application information**
+   with the new IP address.
+
+Configure your JupyterHub to use the GitHub Oathenticator
```
betatim: Oathenticator -> Oauthenticator

choldgraf:
Just rebased so the integration tests should pass and there aren't any more conflicts. Also fixed @betatim 's typo spot.

choldgraf:
Also split up the docs requirements files into two files so that we can get the docs tests to pass

yuvipanda:
@choldgraf tyvm for the patch! I've made two changs: rename 'envt' to 'env', and remove the FAQ section. I don't think the sudo -E explanation belongs in FAQ, but I can't figure out a nice place to put it. So instead i'm just going to remove requiring sudo -E for most things in a different PR. Hope that's ok!

yuvipanda:
#135 removes -E from all tljh-config commands. I'll think about how to remove it from conda and pip commands.

yuvipanda:
@choldgraf just a quick note - I see you had added 'sudo -E' for apt, mkdir etc too in this PR. Those are not necessary. The only commands -E is needed for is tljh-config, the pip and conda we install into the user environments. I've fixed these up as part of #135

choldgraf:
Cool! No worries. I was putting sudo - E in front of everything just for consistency. It seemed confusing to tell people to only use it sometimes and not other times. If it's used rarely enough, I don't think it's a big dealt tho

PR #125, opened 2018-08-10
Update bootstrap.py to require Python 3.6

rgerkin:
The `installer.py` file has f-strings, which are only supported in Python 3.6+.  So installation fails with e.g. Python 3.5.  The boostrap.py file has no version requirements and will try to install TLJH even if it apt installs Python 3.5 when asked for python3.  Technically any Ubuntu 18.04+ probably just installs 3.6, but certainly 16.04 does not without adding special repositories.  One way or another, there should be an explicit check for 3.6+.

### DIFF
--- a/bootstrap/bootstrap.py
+++ b/bootstrap/bootstrap.py
@@ -8,7 +8,7 @@ This script is run as:
     curl <script-url> | sudo python3 -
 
 Constraints:
-  - Be compatible with Python 3.4 (since we support Ubuntu 16.04)
+  - Be compatible with Python 3.6 (since we support Ubuntu 16.04)
   - Use stdlib modules only
 """
 import os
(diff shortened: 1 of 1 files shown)


rgerkin:
I think there are other `python3` commands elsewhere in the installer, causing the integration tests to fail, although if someone (like me) does have an existing python3 installation pointing to 3.5, the TLJH installation still works if those `python3` commands don't try to do 3.6+ requiring stuff.  

Overall, the right approach to making sure that 3.6+ is used everywhere would be to apt install 3.6 and then alias it to python3-tljh or something, and use that everywhere in place of python3 to avoid conflicts with any system python3 that could be of a lower version.

yuvipanda:
Apologies for the late response!

We decided to *not* support 16.04, and require 18.04 for now primarily because of this. python3.6 is not available from the repos for 16.04 (https://packages.ubuntu.com/search?keywords=python3.6&searchon=names&suite=all&section=all reports that it was added only in 17.10). 

There's an open issue in #2 to actively error out if system python is not 3.6. 

Hope this helps!

PR #126, opened 2018-08-11
removing extra copybutton files

choldgraf:

I just realized that we had an old CSS and JS file from before the "sphinx-copybutton" plugin was a thing. This removes those files, and also updates the docs to use the same readthedocs config that binderhub uses (mostly). I'm gonna self merge this tomorrow AM if the tests pass since it may take some tweaking to get RTD to work properly :-/


docs/environment.yml
```diff
@@ -0,0 +1,10 @@
+name: tljh
+channels:
+  - conda-forge
+dependencies:
+- python=3.6
+- traitlets>=4.1
+- sphinx>=1.4, !=1.5.4
+- pip:
+  - recommonmark==0.4.0
```
yuvipanda: we don't need recommonmark :)

yuvipanda:
@choldgraf I'll leave this to you to self merge and iterate on!

docs/environment.yml
choldgraf: oops! ok removing that line and merging...I may merge a couple more to get RTD working if something is off

yuvipanda:
@choldgraf I do! Did the entire theme change?!

choldgraf:
yeah I dunno what's up with the theme changing, it happened with the other jupyter projects too. I wonder if this is an update to Alabaster?

yuvipanda:
@choldgraf yeah, I *think* so. I love the new theme's typography! Except it's cutting the title of the project awkwardly :) Otherwise <3

choldgraf:
haha yeah I saw that too, but I agree I think it is cleaner. We can always reduce the H1 size (or whatever that element is) by a little bit :-)

PR #127, opened 2018-08-11
Add plugin support to the installer

yuvipanda:
TLJH is a great base to build 'stacks' on top of
for various use cases. These 'stacks' should be built by
people who are domain experts in their fields, and easily
updateable with new TLJH versions. Extension points need
to be very clearly defined & evolvable, so we can modify
TLJH without fear of breaking everything.

[pluggy](https://pluggy.readthedocs.io/) is the plugin
mechanism for pytest spun out into its own library,
and fits our requirements well.

There is an experimental pangeo stack in progress at
https://github.com/yuvipanda/tljh-pangeo for an example
of how this would work

 - [x] Add tests

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,10 +68,16 @@ jobs:
                     .circleci/integration-test.py build-image
 
             - run:
-                name: Run basic tests tests
+                name: Run basic tests
                 command: |
                     .circleci/integration-test.py run-test basic-tests test_hub.py test_install.py test_extensions.py
 
+            - run:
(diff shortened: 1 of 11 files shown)


yuvipanda:
Documentation needs two parts:

1. Specifying `--plugin` to the installer
2. How to write a plugin

Will have minimal skeletons of these up before merging.

yuvipanda:
Ok, somewhat reasonable documentation exists for this now! #133 and #132 would make things nicer, but this is good to go! <3

PR #129, opened 2018-08-11
Make it easier to run multiple independent integration tests

### LINKED ISSUE #97: Run multiple integration tests independent of each other
Currently our integration tests are all run on the same container. This provides no isolation - results of one integration test will affect every other one.

Instead, we should run the installer once, commit the new image and use that to run each test. Much more legit.

We can't run the installer in the Dockerfile, since the `systemctl start` commands used by the installer will not work without a running systemd.

yuvipanda:
Provide higher level run-test command in integration-tests.
This runs a number of pytest files in the same container
with the same installation.

 - [x] Add tests

Fixes #97

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,12 +65,12 @@ jobs:
             - run:
                 name: build systemd image
                 command: |
-                    python3 .circleci/integration-test.py build-image
+                    .circleci/integration-test.py build-image
 
             - run:
-                name: Run hub tests
+                name: Run basic tests tests
                 command: |
-                    python3 .circleci/integration-test.py run-test test_hub.py
+                    .circleci/integration-test.py run-test basic-tests test_hub.py test_install.py test_extensions.py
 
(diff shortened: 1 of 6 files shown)


yuvipanda:
I'll add docs on how you can run integration tests locally before merging.

consideRatio:
Ohhh I noticed some `systemd` references in the code! I have lots to learn about Linux still but I'm trying to learn everything I can about these things now. `minikube` requires `kubeadm` to bootstrap k8s 1.11 clusters (`--bootstrapper localkube` is no longer supported), and kubeadm requires systemd, and systemd in a container may require privileged container execution, and that may require CircleCI to first start a VM that in turns start the privileged container rather than directly starting a container. Yikes!

Very curious to learn more about the CI you setup for TLJH, this is quite complicated and I have lots to learn! Right now I'll get back to updating docs and preparing release of z2jh pre JupyterCon.

PR #124, opened 2018-08-10
adding copy button to code blocks and fixing the integration bug

choldgraf:
- [X] Add / update documentation

This PR does two things:

* adds the sphinx copybutton to the code blocks
* fixes the bug that was causing integration tests to fail.
    * For some reason the result of loading the yaml was no longer an instance of `dict` but instead was an instance of `CommentedMap`. That was causing [this line](https://github.com/jupyterhub/the-littlest-jupyterhub/blob/master/tljh/config.py#L53) to always return True, which was creating an empty dictionary and removing any previous values in that final node. This PR checks for that type as well and it's now passing tests.

closes #122

PR #130, opened 2018-08-12
Switch back to a venv for docs + fix .circle config

yuvipanda:
- [x] Add tests

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -104,7 +104,7 @@ jobs:
                 name: install dependencies
                 command: |
                     pip install sphinx
-                    pip install -r docs/doc-requirements.txt
+                    pip install -r docs/requirements.txt
 
             - save_cache:
                 paths:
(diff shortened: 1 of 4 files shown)


yuvipanda:
Looks like there was a conflict in #124, which seems to have added an environment.yml file for docs and maybe removed the docs-requirement.txt file that the documentation integration test was looking for. Since there wasn't anything hard to do with pip in the environment.yml file, I've moved it back to requirements.txt file since it's easier to do that than try to convert the integration test container to have conda :)

choldgraf:
yo - I mostly copied over the environment.yml file because it was used in the other jupyterhub projects, and since it was the only way I could figure out how to get RTD builds working (but perhaps we can point the `.readthedocs` config file to the `pip` file instead). I'll try to take a look at it on my flight.

choldgraf:
Also - note that my intention was actually to use both environment.yml and a requirements.txt file. It's just that this is wrapped up in the still-unmerged PR from Zexuan's user test:

https://github.com/jupyterhub/the-littlest-jupyterhub/pull/123/files#diff-b0278a395ac6448087984d5186dbe855R9

PR #134, opened 2018-08-13
[WIP] adding uninstallation script

### LINKED ISSUE #128: Add a command to uninstall TLJH
I've had a few moments where I need to delete the hub and just start fresh. I know to delete `/opt/tljh`, but I always feel like I might be missing something. Moreover, most people probably won't think to check `/opt/tljh`. What if we added a `tljh-config clean` command that would require you to confirm, and if so would completely erase all jupyterhub-related changes to the machine?

choldgraf:
 - [ ] Add tests

 This PR is meant to add an "uninstall" argument to the `tljh-config` tool. Here are some open questions:

1. Can it *really* delete everything without breaking itself in the middle of deleting things?
2. Can it delete the JupyterHub without really confusingly breaking someone's session if they're typing from a Jupyter Notebook server terminal (or jupyter lab).
3. Is the only thing we'd need to delete the `/opt/tljh` folder?
4. Should user directories be deleted by default

Depending on the answers to 1 and 2, I may or may not be +1 on this feature being added :-)

closes #128 if it's merged

### DIFF
--- a/tljh/config.py
+++ b/tljh/config.py
@@ -192,6 +192,21 @@ def reload_component(component):
         print('Proxy reload with new configuration complete')
 
 
+def uninstall_tljh(confirm):
+    """Uninstall TLJH entirely from a machine."""
+    if confirm is False:
+        print('To confirm that you wish to uninstall JupyterHub entirely, '
+              're-run this command with the text "confirm", like so: \n\n'
+              '     tljh-config uninstall confirm')
+        return
+
+    # Uninstall all JupyterHub components
+    os.removedirs('/opt/tljh')
+    cmd = "curl https://raw.githubusercontent.com/jupyterhub/the-littlest-jupyterhub/master/bootstrap/bootstrap.py | sudo python3 - --admin <admin-user-name>"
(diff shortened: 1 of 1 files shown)


yuvipanda:
Thanks for opening this!

I don't know the answer to (1) :)

Answer to (2) is most definitely shit is gonna fuck up in weird ways.

https://the-littlest-jupyterhub.readthedocs.io/en/latest/topic/installer-actions.html has answers to (3). Short answer is 'a lot more'.

(4) is for sure no.

I'd also like this to be tljh-uninstall, in a file of its own rather than rolled into tljh-config (which should only do configuration related things). 

It's also unclear how this should interact with the plugin mechanisms added in #127.

choldgraf:
I'm gonna close this, because we decided in #128 that it'd be better to capture this use-case with good documentation rather than a one-click solution!

ptiwari2407:
@choldgraf , you closed the previous one by pointing to this and now you close this one and pointing to another thread. This is ridiculous. Where is the right solution to uninstall?

PR #137, opened 2018-08-13
Added existing property-path for tljh-config set method

ynnelson:
Updated topic/tljh-config.rst to include some of the property paths that are declared in the code base but not yet documented.


ynnelson:
This was just an edit of the documentation that I think will be useful if anyone can review/merge.

minrk:
Great!

PR #138, opened 2018-08-13
Use node 8, not 10

yuvipanda:
JupyterLab currently doesn't let you install extensions
with node 10, so move back to 8


yuvipanda:
@ian-r-rose ah, ok. Then maybe I'll abandon this :D LMK when the release goes out?

yuvipanda:
Going to actually merge this, since I'm at the PANGEO workshop and we wanna see if we can make a nice pangeo stack plugin.

PR #153, opened 2018-08-23
Don't set a memory limit by default

yuvipanda:
Users should set up their own memory limits after install.
Better than a surprise when they try using more than 1G.

 - [ ] Add tests

### DIFF
--- a/tljh/configurer.py
+++ b/tljh/configurer.py
@@ -28,7 +28,7 @@ default = {
         'admin': [],
     },
     'limits': {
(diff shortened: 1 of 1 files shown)


minrk:
[Relevant docs](https://the-littlest-jupyterhub.readthedocs.io/en/latest/topic/tljh-config.html?highlight=limit#set-a-configuration-property) that might want updating. Maybe mention that there's no limit by default?

minrk:
Finished in #202

PR #157, opened 2018-08-27
Draft tool to change password when using Firsttimeauthrenticator

Carreau:
By default TLJH uses first time authenticator, and obviously users may
forget their passwords. This provide a base as to how a CLI tools may be
provided to allow a root user to update a password for a given user

 - [ ] Add tests

Things to add on top of this:
 - Documentation
 - Run it with the right python. 

Is there any thoughts of providing a small `tljh` command that can be use to do basic adminitration of `tljh` ?

### DIFF
--- a/tools/passwd.py
+++ b/tools/passwd.py
@@ -1,0 +1,17 @@
+#!/usr/bin/env python3
+# hub bin python3
+import dbm
+from getpass import getpass
+
+import bcrypt
+
+install_prefix = os.environ.get('TLJH_INSTALL_PREFIX', '/opt/tljh')
+
+if __name__ == '__main__':
+    print('create user or update password:')
+    username = input('Username: ').encode()
+    password = getpass()
+    with dbm.open(install_prefix+'/state/passwords.dbm', 'c', 0o600) as db:
+        db[username] = bcrypt.hashpw(password.encode(), bcrypt.gensalt())
+    print('password for user `{username}` created or updated'.format(username=username))
+


yuvipanda:
This is a great start! However, I believe this tool should be in https://github.com/yuvipanda/jupyterhub-firstuseauthenticator and not here. 

Thank you for working on it!

yuvipanda:
It would be great for us to figure out how to 'shim' commands under tljh - much the same way we do for 'jupyter'. So this could be tljh firstuseauth reset-user <username> or something.

Carreau:
> This is a great start! However, I believe this tool should be in yuvipanda/jupyterhub-firstuseauthenticator and not here.

I agree with that, but I'm unsure it is really practical (in this case), as it's going to be complex to carry the informations of where the hub is running/where the dbm file is, without spawning a full app; and as you said we need to find a way to shim commands (we can likely use entry points if things are Python packages. 

Would that be ok if (for now) I move the code in FUA, and conditionally import it in `tljh-config` ?

yuvipanda:
I don't really want us to have any specific auth related code here, since once we add it it's going to be pretty impossible to remove. This is why I want us to have a system to do 'tljh <something>' - where we can, say, pass environment variables to <something> for it to know where things are?

@Carreau maybe we can start this by making it a [plugin](https://the-littlest-jupyterhub.readthedocs.io/en/latest/contributing/plugins.html)? This way we can unblock people right now, and try to figure out what to do without the strong time pressure.

minrk:
What about making the entrypoint for the plugin in the firstuseauthenticator package itself? Seems simpler than maintaining/distributing another package.

yuvipanda:
Superseded by #175

PR #160, opened 2018-08-28
jupyterhub 0.9.2

minrk:
fixes TERM shutdown

bumps oauthenticator to 0.8 as well


minrk:
Turns out we were actually *relying* on the bug in 0.9.1, where cleanup wouldn't fire due to an error in the SIGTERM handler. This is fixed by setting `c.JupyterHub.cleanup_servers = False` in config.

yuvipanda:
Yay tests that catch things!!!! <3

PR #161, opened 2018-08-28
test manual https setup

### LINKED ISSUE #92: Add integration test for HTTPS

minrk:
adds integration test for manual https certs

closes #92

- [x] Add tests

### DIFF
--- a/integration-tests/test_proxy.py
+++ b/integration-tests/test_proxy.py
@@ -1,0 +1,61 @@
+"""tests for the proxy"""
+import os
+import shutil
+import ssl
+from subprocess import check_call
+import time
+
+import requests
+
+from tljh.config import reload_component, set_config_value, CONFIG_FILE
+
+
+def test_manual_https(preserve_config):
+    ssl_dir = "/etc/tljh-ssl-test"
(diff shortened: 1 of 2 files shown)


integration-tests/test_proxy.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def test_manual_https(preserve_config):
+    ssl_dir = "/etc/tljh-ssl-test"
+    key = ssl_dir + "/ssl.key"
+    cert = ssl_dir + "/ssl.cert"
+    os.makedirs(ssl_dir, exist_ok=True)
+    os.chmod(ssl_dir, 0o600)
+    # generate key and cert
+    Popen(
```
yuvipanda: Why Popen rather than check_output? Since we aren't calling 'wait' explicitly here there's a race condition here too

minrk: 👍

yuvipanda:
Thanks, @minrk

PR #162, opened 2018-08-28
missing arguments in integration test commands

minrk:
and missing start-container command

integration-test was refactored a bit, and some of the commands not used on circleci stopped working

Having these helps for local testing


yuvipanda:
Thanks, @minrk

PR #163, opened 2018-08-28
put config in `$tljh/config` directory

minrk:
easier to maintain proper private permissions and consolidate, since we have multiple config files (config.yaml, jupyterhub_config.d)

I thought there was an issue about this, but I can't seem to find it right now.

- [x] Add tests


yuvipanda:
@minrk we should also probably write a migrator?

minrk:
Yeah, we probably should have a migrator at some point. It seems a bit early to do it right now, though.

yuvipanda:
@minrk unfortunately I think there's enough people already using this for the fall classes that we should do a migrator...

Also we should probably have a migrator / backwards compat test...

minrk:
Migration was already implemented and tested, so I moved those to a `migrator.py` to make it easier to add new migrations.

PR #167, opened 2018-08-29
installer.py: remove unused f"..."

gyg-github:
removed formatting  symbol f'', since "....x" is considered as a type in the formatting.
actual directory  is just "https://github.com/nodesource/distributions/blob/master/deb/setup_8.x"

 - [ ] Add tests


minrk:
Thanks!

Issue #158, opened 2018-08-28
check/change default umask


Carreau:
While walking in central valley today I came across a wild jupyterhub running in a grass field. It was used to teach R. Admins were surprised that students could see each other files.

Would it make sens to me to change the umask to 025 (I believe), use the `-M 750` of `adduser` when possible (not available on Ubuntu AFAICT), or at least `subprocess.check_call(['chmod','0750', f'~{user}'])`.

It seem to be a reasonable default IMHO

yuvipanda:
This was fixed in #169

PR #169, opened 2018-08-29
Don't create home publicly readable

Carreau:
World-Readable seem to be a surprising default for many people,
especially in teaching context. Switch to a more reasonable `rwxr-x---`

We have to issue a chmod, as changing at creation time would require
changin /etc/adduser.conf DIR_MODE=0760 (or whatever), but that seem
unwise.

We do not set the exact permission in case the DIR_MODE is more
restrictive.

Closing #158

 - [x] Add tests

(cc @mstobb)

One question is do we want that to be configurable ? I tend to think no.

### DIFF
--- a/docs/topic/security.rst
+++ b/docs/topic/security.rst
@@ -22,6 +22,10 @@ permissions.
 
 #. A home directory is created for the user under ``/home/jupyter-<username>``.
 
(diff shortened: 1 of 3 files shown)


Carreau:
Circle is happy !

minrk:
Nice!

PR #179, opened 2018-09-12
Normalize systemuser

### LINKED ISSUE #146: [bug] Need to check for linux username acceptability
I had a user with a particularly long email address which they used to login to TLJH (we use Auth0 for authentication).  When TLJH tried to create their account, it gave them a 500 Server Error.  According to journalctl: 

`subprocess.CalledProcessError: Command '['useradd', '--create-home', ‘jupyter-user.name.replaced@gmail.com']' returned non-zero exit status 3.`

I discovered that `sudo -E useradd --create-home jupyter-user.name.replaced@gmail.com` fails with `useradd: invalid user name 'jupyter-user.name.replaced@gmail.com'` but shorter usernames work.  Turns out the useradd man page states "Usernames may only be up to 32 characters long.".  So the problem is TLJH is making no attempt to enforce this limit.  

So there are two approaches I can think of to properly handling these longer names (I make no claim to viability of either):

1. Present a warning to the user to choose a shorter username (shorter than 24 characters would allow 'jupyter-' prefix) and have the admin add it.
2. Massage the user's linux username to be different than their TLJH login credentials so that the linux username is acceptable.  Not sure how doable this is.

yuvipanda:
 - [x] Add tests

Fixes #146

### DIFF
--- a/docs/topic/security.rst
+++ b/docs/topic/security.rst
@@ -18,7 +18,10 @@ permissions.
 #. The unix user account created for a JupyterHub user named ``<username>`` is
    ``jupyter-<username>``. This prefix helps prevent clashes with users that
    already exist - otherwise a user named ``root`` can trivially gain full root
-   access to your server.
+   access to your server. If the username (including the ``jupyter-`` prefix)
+   is longer than 26 characters, it is truncated at 26 characters & a 5 charcter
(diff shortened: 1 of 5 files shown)


JuanCab:
Thanks!

PR #205, opened 2018-10-18
updating theme and storing docs artifacts

choldgraf:
- [X] Add / update documentation
 
This updates the documentation theme to the common jupyterhub one and stores artifacts as part of circle. It also deletes `doc-requirements.txt` because I think it wasn't being used at all

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,2 +1,4 @@
 sphinx>=1.4, !=1.5.4
 sphinx_copybutton
+alabaster
+alabaster_jupyterhub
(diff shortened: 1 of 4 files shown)


choldgraf:
The built documentation: https://891-137537346-gh.circle-artifacts.com/0/html/index.html

docs/requirements.txt
```diff
@@ -1,2 +1,4 @@
 sphinx>=1.4, !=1.5.4
 sphinx_copybutton
+alabaster
```
willingc: Alabaster is unneeded as a dependency as both sphinx and alabaster_jupyterhub would install. @choldgraf @betatim

PR #209, opened 2018-10-21
add warning if tljh-config is called as non-root user

### LINKED ISSUE #120: Warn & abort if tljh-config is called as non-root user
Is there a way to quickly double check that:

1. Sudo is invoked (e.g. by trying to do something that needs sudo)
2. -E was used (e.g. by trying to access a variable we know only exists in the user's environment)

and if one of these conditions isn't met, throw a helpful error telling them to use `sudo -E`?

anyushevai:
There used to be two separate errors that occurred if we run the command without sudo:

[Errno 13] Permission denied: '/opt/tljh/installer.log'
at line 226

And 
[Errno 13] Permission denied: '/opt/tljh/config/config.yaml'
at lines 294-305

Now we print an error and print a message about using 'sudo -E'.

I thought about putting the whole main function into one try-except block, but it seems more appropriate to use two different try-except blocks.

https://github.com/jupyterhub/outreachy/issues/5

closes #120

### DIFF
--- a/tljh/config.py
+++ b/tljh/config.py
@@ -223,7 +223,11 @@ def main(argv=None):
         argv = sys.argv[1:]
 
     from .log import init_logging
-    init_logging()
+    try:
+        init_logging()
+    except Exception as e:
+        print(str(e))
+        print("Perhaps you didn't use `sudo -E`?")
 
     argparser = argparse.ArgumentParser()
     argparser.add_argument(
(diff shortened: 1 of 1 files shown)


tljh/config.py
```diff
@@ -291,18 +295,22 @@ def main(argv=None):
...
+            set_config_value(args.config_path, args.key_path, parse_value(args.value))
+        elif args.action == 'add-item':
+            add_config_value(args.config_path, args.key_path, parse_value(args.value))
+        elif args.action == 'remove-item':
+            remove_config_value(args.config_path, args.key_path, parse_value(args.value))
+        elif args.action == 'reload':
+            reload_component(args.component)
+        else:
+            argparser.print_help()
+    except Exception as e:
```
minrk: Rather than catching all errors here, let's try proactively checking `if os.geteuid() != 0`, which is the condition when it is run as a non-root user. Maybe show an early warning in this case, indicating that it may fail with permission errors?

tljh/config.py
```diff
@@ -219,11 +219,19 @@ def _is_list(item):
 
 
 def main(argv=None):
+    if os.geteuid() != 0:
+        print("Perhaps you didn't use `sudo -E`?")
```
choldgraf:
```suggestion
        print("It looks like this command wasn't run with root privileges. Perhaps you didn't use `sudo -E`?")
```

tljh/config.py
```diff
@@ -219,11 +219,19 @@ def _is_list(item):
 
 
 def main(argv=None):
+    if os.geteuid() != 0:
+        print("It looks like this command wasn't run with root privileges. Perhaps you didn't use `sudo -E`?")
+        raise(SystemExit)
```
minrk: Instead of exiting, let's just print this warning. That way, people will see the message *if* they hit an error, but if there is no error, it's okay, too.

```suggestion

```

choldgraf:
This LGTM! If @minrk is +1 on the way to check for root then I'm +1 on merge! Thanks @anyushevai 🎉

minrk:
Thanks, @anyushevai!

PR #212, opened 2018-10-25
Update jupyterhub_config.py

muralikarthikk:
 - [ ] Add tests

### DIFF
--- a/tljh/jupyterhub_config.py
+++ b/tljh/jupyterhub_config.py
@@ -34,7 +34,7 @@ class UserCreatingSpawner(SystemdSpawner):
         else:
             user.remove_user_group(system_username, 'jupyterhub-admins')
         return super().start()
-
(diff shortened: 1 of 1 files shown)


tljh/jupyterhub_config.py
```diff
@@ -34,7 +34,7 @@ def start(self):
         else:
             user.remove_user_group(system_username, 'jupyterhub-admins')
         return super().start()
-
+c.NotebookApp.notebook_dir = '/home/efs/'
```
minrk: I don't think this is what we want for the default behavior. Did you mean to update your own configuration?

sosiristseng:
Perhaps we should discard this PR since it looked like a misclick and the source of the PR is already gone.

yuvipanda:
Feel free to re-open if needed, @muralikarthikk

PR #213, opened 2018-10-25
remove config values

gillybops:
Remove config values in the-littlest-jupyterhub by editing 'config.py' file
and making the change. ie. comment out or delete the value to be unset

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -123,6 +123,10 @@ jobs:
                     cd docs
                     make html
 
+            - store_artifacts:
+                path: docs/_build/html/
+                destination: html
+
 workflows:
     version: 2
     all-tests:
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -34,7 +34,10 @@ intersphinx_cache_limit = 90 # days
 # The name of the Pygments (syntax highlighting) style to use.
 pygments_style = 'sphinx'
 
-html_theme = 'alabaster'
+import alabaster_jupyterhub
+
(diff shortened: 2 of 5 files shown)


minrk:
Thanks for the PR! I think we also want to do this a little differently. Instead of creating a new file, this should be added as a new section to `docs/topic/tljh-config.rst`. See [ReStructuredText](http://docutils.sourceforge.net/docs/user/rst/quickstart.html) for reference on how to write rst.

Additionally, there is not a `config.py`, but there is a `config.yaml`. It would be useful to also mention `tljh-config remove-item`:

```
tljh-config -h
usage: tljh-config [-h] [--config-path CONFIG_PATH]
                   {show,set,add-item,remove-item,reload} ...

positional arguments:
  {show,set,add-item,remove-item,reload}
    show                Show current configuration
    set                 Set a configuration property
    add-item            Add a value to a list for a configuration property
    remove-item         Remove a value from a list for a configuration    <---
                        property
    reload              Reload a component to apply configuration change

optional arguments:
  -h, --help            show this help message and exit
  --config-path CONFIG_PATH
                        Path to TLJH config.yaml file
```

You can update this PR by pushing new commits to your `master` branch.

gillybops:
Section 'Remove configuration values' added below 'Advanced: config.yaml' section of  tljh-config.rst

fm75:
I ended up editing the `yaml` file by hand, as well. I never was able to get the "remove-item" syntax correct. Maybe there should be an example, especially since the `yaml` files are kind of unformatted.

Issue #206, opened 2018-10-19
Unable to install on Ubuntu 14.04


xtutran:
Hi,

I am facing on this error on Ubuntu 14.04 

This is error when I was trying to use ``bootstrap.py``
```
Upgrading TLJH installer...
Traceback (most recent call last):
  File "bootstrap.py", line 83, in <module>
    main()
  File "bootstrap.py", line 68, in main
    ] + pip_flags + [tljh_repo_path], stderr=subprocess.STDOUT)
  File "/usr/lib/python3.4/subprocess.py", line 620, in check_output
    raise CalledProcessError(retcode, process.args, output=output)
subprocess.CalledProcessError: Command '['/opt/tljh/hub/bin/pip', 'install', '--upgrade', 'git+https://github.com/jupyterhub/the-littlest-jupyterhub.git']' returned non-zero exit status 1
```

This is something when I was trying to run ``setup.py``

``
sudo /opt/tljh/hub/bin/python3.4 setup.py install
``

Error message
``
error in the-littlest-jupyterhub setup command: 'install_requires' must be a string or list of strings containing valid project/version requirement specifiers
``

This is a little bit strange to me. Do you guys have any idea?

Thank you,
Tran

minrk:
the-littlest-jupyterhub requires Ubuntu >= 18.04. My guess is the errors come from setuptools and/or Python being too old (Python 3.4 will not work).

choldgraf:
@minrk perhaps we should add a short "requirements" section to the index page? https://the-littlest-jupyterhub.readthedocs.io/en/latest/

betatim:
It is mentioned here https://the-littlest-jupyterhub.readthedocs.io/en/latest/#installation

choldgraf:
yeah, just a subsection for "requirements" even if it's followed by just that sentence would make things more explicit IMO

willingc:
Why not put it in an important directive?

xtutran:
> the-littlest-jupyterhub requires Ubuntu >= 18.04. My guess is the errors come from setuptools and/or Python being too old (Python 3.4 will not work).

Hi,

Thanks for the info! I just realised that there might more issues with Ubuntu 14.04. So I think I will give another try in Ubuntu 18.04

Thanks

choldgraf:
hey @xtutran - does this seem to have worked for you? :-)

yuvipanda:
[...]

PR #221, opened 2018-10-31
Provide better error message when running on unsupported distro

### LINKED ISSUE #186: Won't install on Ubuntu 16.04
I'm trying to install tljh in Ubuntu 16.04, but I get the following error:

    Checking if TLJH is already installed...
    TLJH already installed, upgrading...
    Upgrading TLJH installer...
    Setup tljh package
    Starting TLJH installer...
      File "/usr/lib/python3.5/runpy.py", line 174, in _run_module_as_main
        mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
      File "/usr/lib/python3.5/runpy.py", line 144, in _get_module_details
        code = loader.get_code(mod_name)
      File "<frozen importlib._bootstrap_external>", line 767, in get_code
      File "<frozen importlib._bootstrap_external>", line 727, in source_to_code
      File "<frozen importlib._bootstrap>", line 222, in _call_with_frames_removed
      File "/opt/tljh/hub/lib/python3.5/site-packages/tljh/installer.py", line 326
        raise FileExistsError(f'/usr/bin/tljh-config exists but is not a symlink to {tljh_config_src}')
                                                                                                 ^
    SyntaxError: invalid syntax

I see that the offending line uses a string feature ('f' prefix) that is not available in python 3.5, which is the most recent python in Ubuntu 16.04.

yuvipanda:
 - [x] Add tests

Fixes #186 #206

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,9 +80,9 @@ jobs:
                         plugins test_simplest_plugin.py
 
             - run:
-                name: Run distro check test
+                name: Run bootstrap checks
                 command: |
-                    py.test integration-tests/test_distro.py
+                    py.test integration-tests/test_bootstrap.py
 
 
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,7 +57,7 @@ jobs:
             - run:
                 name: setup python3
                 command: |
-                    apk add --no-cache python3
+                    apk add --no-cache python3 pytest
 
             - checkout
             - setup_remote_docker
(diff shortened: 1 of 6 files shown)


yuvipanda:
This provides a better error message, and bolder text about requiring Ubuntu 18.04.

Carreau:
I believe platform will be removed from stdlib in 3.8. I need to check.

choldgraf:
ahh yea, on

https://docs.python.org/2/library/platform.html#platform.linux_distribution

I see:

```
Note This function is deprecated since Python 3.5 and removed in Python 3.8. See alternative like the distro package.
```

looks like the usage for "distro" is pretty similar:

https://pypi.org/project/distro/

yuvipanda:
Unfortunately we have to only use stdlib in bootstrap.py. I'll probably switch it to read from /etc/os-release. Thanks for the pointer, @carreau and @choldgraf

yuvipanda:
@Carreau @choldgraf I've modified this to use /etc/os-release now!

choldgraf:
is this ready to go now?

Carreau:
Awayyyyy we go.

PR #224, opened 2018-11-01
consolidate yaml configuration

minrk:
workaround ruamel.yaml issue https://bitbucket.org/ruamel/yaml/issues/255, where once an empty dict or list has been written, 'flow' style is used thereafter, using dense `{key: value}` form instead of traditional yaml block style.

 - [x] Add tests

### DIFF
--- a/tests/test_yaml.py
+++ b/tests/test_yaml.py
@@ -1,0 +1,19 @@
+from tljh.yaml import yaml
+
+
+def test_no_empty_flow(tmpdir):
+    path = tmpdir.join("config.yaml")
+    with path.open("w") as f:
+        f.write("{}")
+    # load empty config file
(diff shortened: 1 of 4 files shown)


choldgraf:
this is something that's only developer-facing right? People are only going to use the `tljh-config` tool to edit yaml files. (I'm trying to figure out if this needs docs or not)

yuvipanda:
This should make the issue @fm75 reported in https://github.com/jupyterhub/the-littlest-jupyterhub/pull/213#issuecomment-436856153 easier.

yuvipanda:
@choldgraf I think this is ok in not needing docs. We can probably remove it when the next release of ruamel.yaml comes out.

Issue #219, opened 2018-10-30
Make bootstrap.py work better with intercepting proxies


fm75:
Issue #11 Addresses the need to document getting the environment to be used by `sudo -E`. I managed to install `TLJH` successfully this morning on an `ubuntu 18.04` instance running in an OpenStack private cloud. Things that I encountered that required it.

- http(s)_proxy settings to get through our firewalls
- SSL error from requests installing `Miniconda3`
- SSL error from `npm`

I believe that these can all be addressed by updating [Installing on your own server](https://the-littlest-jupyterhub.readthedocs.io/en/latest/install/custom-server.html)

Before the `sudo apt-get install python3 git curl`

    export http_proxy=<your_proxy>
    export https_proxy=<your_proxy>
    export REQUESTS_CA_BUNDLE=/etc/ssl/certs/ca-certificates
    sudo npm config set strict-ssl false

and then insert " -E " in the pipe to python3:
```bash
curl https://raw.githubusercontent.com/jupyterhub/the-littlest-jupyterhub/master/bootstrap/bootstrap.py | sudo -E python3 - --admin <admin-user-name>
```

fm75:
I have a [documentation only change for this](https://github.com/fm75/the-littlest-jupyterhub/blob/documentation-behind-proxy/docs/install/custom-server.rst). Please let me know if you would like a PR for it.

choldgraf:
hey @fm75 - thanks for catching this! Documentation improvements are *always* welcome! Do you see any downsides to including these changes as part of the default "on your own server" TLJH install? 

One thing I wonder is whether there should be another section to the custom server install page, called something like "common gotchas" where we can list things like configuring SSL properly etc. What do you think?

fm75:
@choldgraf - Thanks. I wouldn't know about 'common gotchas'. I think the "custom server install" opens up a range of things that the cloud solutions don't have because they have made choices on a number of configuration options. 

### Troubleshooting
There is a "chapter" in the docs for Troubleshooting. Currently it only has:
- Looking at logs. 
  - (Perhaps this might be improved. I felt that I need to understand more about how the logging was working to investigate some stuff. That said, I think the stuff that I had the biggest problem with was because I tried to delete a user, then tried to add him back - long story)
- Google Cloud
- Amazon Web Services.

I suppose adding a section for Custom Server would be useful.

### With a firewall
- Need proxy settings to get out for the installation.
[...]

PR #230, opened 2018-11-08
Documentation behind proxy

fm75:
 - [ ] Add tests

This addresses issue #219.
Perhaps you won't consider this within the existing requirements because it does not strictly meet Pre-requisite 4 - a **Public IP** address. If we relax the *public* to mean that it is reachable from your audience, which is on the same private side of a proxy or proxies, then I think this is a good modification.

I thought about offering several choices for a one-line installer something like:
```bash
http_proxy=http://your.domain.here:port curl https://raw.githubusercontent.com/jupyterhub/the-littlest-jupyterhub/master/bootstrap/bootstrap.py | sudo python3 - --admin <admin-user-name>
```
but exporting the proxy information will be needed by the installer once bootstrap runs it.

Is there a better answer for the `npm` issue than just going insecure?
Any suggestions welcome.


docs/install/custom-server.rst
```diff
@@ -31,11 +31,25 @@ Step 1: Installing The Littlest JupyterHub
 #. Using a terminal program, SSH into your server. This should give you a prompt where you can
    type commands.
 
+#. If your server is behind a firewall and needs a proxy:
```
yuvipanda: Can we move this to the 'troubleshooting section' under 'your own server' (like we have for GCP and AWS)? We can link to that more prominently here as well. I'd rather we do that than put this in here explicitly.

docs/install/custom-server.rst
```diff
@@ -31,11 +31,25 @@ Step 1: Installing The Littlest JupyterHub
...
+
+   .. code-block:: bash
+
+      export http_proxy=<your_proxy>
+      export https_proxy=<your_proxy>
+
+#. Some requests will fail if your certs are self-signed:
```
yuvipanda: Do you have a sense of why this is? Fiddling with CA stores is always dangerous, and I do not want us to do that unless we explicitly need it.

docs/install/custom-server.rst
```diff
@@ -31,11 +31,25 @@ Step 1: Installing The Littlest JupyterHub
 #. Using a terminal program, SSH into your server. This should give you a prompt where you can
    type commands.
 
+#. If your server is behind a firewall and needs a proxy:
+
+   .. code-block:: bash
+
+      export http_proxy=<your_proxy>
```
yuvipanda: Was this all that was needed?

docs/install/custom-server.rst
yuvipanda: I think this can be a section called 'running behind a corporate proxy' or some such, to prevent confusion with running behind a reverse proxy like nginx or apache.

docs/install/custom-server.rst
yuvipanda: I *think* this is because your organization intercepts HTTPS with some sort of MITM to inspect all traffic in and out. Do you know if that is the reason? Can you tell me what error you got and who the 'self' in self-signed is?

fm75:
@yuvipanda Thanks for the review. I am sure you are busy with many other projects and I have appreciated your assistance.

When I get home, I will look at moving the issues to a new section of troubleshooting, as suggested. The issues in the two cloud solutions are about restarting problems, where mine are about installation problems.

If we like the issues there, I would still suggest a "red flag or warning" in the installation instructions that says something like, "if you need to go through a proxy" to reach github/PyPI, ... see the trouble-shooting section, first - so the user does not have to wait until it fails to learn about how to set up the proxy.

Some time in the next few days, I plan to reuse a different VM and will re-run this exercise, so that I can experiment with whether both http_proxy and https_proxy where needed.  Does anybody know how `npm` is different w.r.t. security?

yuvipanda:
> The issues in the two cloud solutions are about restarting problems, where mine are about installation problems.

I think that's ok, hopefully those sections will grow to include more troubleshooting tips in the future!

> If we like the issues there, I would still suggest a "red flag or warning" in the installation instructions that says something like, "if you need to go through a proxy" to reach github/PyPI, ... see the trouble-shooting section, first - so the user does not have to wait until it fails to learn about how to set up the proxy.

This sounds good to me. We can also possibly put this warning in bootstrap.py itself, if we can find ways to detect it.

> Some time in the next few days, I plan to reuse a different VM and will re-run this exercise, so that I can experiment with whether both http_proxy and https_proxy where needed. 

This is awesome and thank you for doing this. I'd count this as two separate things - HTTP / HTTPS proxy, and the certificates on the machine. 

>  Does anybody know how npm is different (from pip/conda) w.r.t. security?

Can you tell me what aspects of security you are referring to?

fm75:
```bash
export REQUESTS_CA_BUNDLE=/etc/ssl/certs/ca-certificates
sudo npm config set strict-ssl false
```
The first line seemed to make the http stuff work. 
I needed the second line or the npm install part failed.

yuvipanda:
```
sudo npm config set strict-ssl false
```

seems extremely dangerous, since it basically turns off HTTPS certificate validation for npm (per https://docs.npmjs.com/misc/config#strict-ssl). I would be somewhat hestitant to recommend anyone do that in TLJH. Do you have any idea from your IT department (or whoever has setup these restrictions) what exactly they have done and why?

fm75:
:) That is why I asked. Researching...

docs/install/custom-server.rst
fm75: I think this must be here. The install simply does not work if it is missing. I will update text per your comment on a firewall type proxy. I can say "corporate", but maybe it is just an "IT" firewall between your server and the internet - even if it is not corporate. LMK on that.

fm75:
@yuvipanda If it is ok with you, I will close this PR, and give you a different one that I think addresses these issues.

But I have two questions before that.
- *A public IP where the server can be accessed from the internet*. (I think this can be relaxed, based on my success.) I am open to suggestions on rewording it
- *This will take 5-10 minutes, and will say 'Done!' ...* This is not currently true. I have added a separate issue for it. Please advise.

yuvipanda:
@fm75 that sounds like a good idea! 

re: 'public IP', perhaps something like 'a way to access the instance from your browser (such as a public IP)'?

fm75:
Ok - will tweak requirement 4 slightly.

fm75:
@yuvipanda - I hope this properly addresses your concerns. Hoping to get this done so I can move onto the trouble-shooting issues.

yuvipanda:
I chatted with @fm75 on Gitter. I'm going to merge this now, but do the move to the troubleshooting section myself later.

Thanks for the PR, @fm75

PR #232, opened 2018-11-13
Exit when tljh-config is called as non-root

yuvipanda:
- Previously, we only printed warning and did not exit
- Split the warning over two lines rather than one
- -E is not required for tljh-config anymore (since it is in
  /usr/bin, not /usr/local/bin), so do not ask users to use it.
  https://github.com/jupyterhub/the-littlest-jupyterhub/pull/135
  is the PR that made this change, and changed documentation in
  most places.

Followup #120 
 - [ ] Add tests

### DIFF
--- a/tljh/config.py
+++ b/tljh/config.py
@@ -219,7 +219,9 @@ def _is_list(item):
 
 def main(argv=None):
     if os.geteuid() != 0:
-        print("It looks like this command wasn't run with root privileges. Perhaps you didn't use `sudo -E`?")
+        print("tljh-config needs root privileges to run", file=sys.stderr)
+        print("Try using sudo before the tljh-config command you wanted to run", file=sys.stderr)
+        sys.exit(1)
 
     if argv is None:
         argv = sys.argv[1:]


minrk:
I think warning and not exit is the right thing to do, since properly configured admin permissions shouldn't require root, only erroring if/when there is an actual permission error. What do you think?

yuvipanda:
Sorry to get back on to this so late!

@minrk I thought we had given up on trying to make this work for admins without root? (https://github.com/jupyterhub/the-littlest-jupyterhub/pull/51)

consideRatio:
I'm not sure what I think would be best with regards to a decisive exit or emitting a warning - I lean towards exit as I value a decisive error message.

No matter on the topic on warning vs error, I think this part is relevant and should not be lost by just closing this PR.

> * -E is not required for tljh-config anymore (since it is in
>   /usr/bin, not /usr/local/bin), so do not ask users to use it.
>   [Move tljh-config symlink to /usr/bin #135](https://github.com/jupyterhub/the-littlest-jupyterhub/pull/135)
>   is the PR that made this change, and changed documentation in
>   most places.

Issue #236, opened 2018-11-22
Installation fails on a fresh 18.10 system (node 8.x has no release candidate)


owah:
Hello community,

Im trying to install the littlest jupyterhub on a fresh Ubuntu Server 18.10 VM. At some point the installer aborts due to:

```
subprocess.CalledProcessError: Command '['apt-get', 'update', '--yes']' returned non-zero exit status 100.
```
When running apt-get update manually, I get:

```
Err:6 https://deb.nodesource.com/node_8.x cosmic Release                 
404  Not Found [IP: 54.230.129.120 443]
Reading package lists... Done                      
E: The repository 'https://deb.nodesource.com/node_8.x cosmic Release' does not have a Release file.
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
N: See apt-secure(8) manpage for repository creation and user configuration details.
```

So I checked out https://deb.nodesource.com/node_8.x/dists/ and there is indeed no release file for cosmic, only for bionic. From node 10.x onwards, there are also cosmic release files.

I assume that this can be fixed by changing the installer to install a newer node version, but I am not sure how much of this project specifically depends on version 8.x, which is why I am not creating a pull request, but rather opening an issue, so that more experienced devs can weigh in on this.

mgd722:
This still fails for me.  Ubuntu 18.04, just updated node to latest, following the custom server install instructions.

yuvipanda:
@mgd722 sorry to hear that! Can you provide a traceback or other error message you get?

mgd722:
Sorry-- should've included that.  When I run:
`curl https://raw.githubusercontent.com/jupyterhub/the-littlest-jupyterhub/master/bootstrap/bootstrap.py | sudo -E python3 - --admin mike`

I get:

```
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  3557  100  3557    0     0  30930      0 --:--:-- --:--:-- --:--:-- 31201
Checking if TLJH is already installed...
Setting up hub environment
Traceback (most recent call last):
  File "<stdin>", line 107, in <module>
  File "<stdin>", line 69, in main
  File "/usr/lib/python3.6/subprocess.py", line 336, in check_output
    **kwargs).stdout
  File "/usr/lib/python3.6/subprocess.py", line 418, in run
    output=stdout, stderr=stderr)
[...]

PR #238, opened 2018-11-23
Update nodejs to 10.x LTS

yuvipanda:
Reverts https://github.com/jupyterhub/the-littlest-jupyterhub/pull/138 and
fixes https://github.com/jupyterhub/the-littlest-jupyterhub/issues/236.


yuvipanda:
I wonder now how this will handle upgrades of TLJH.

yuvipanda:
It looks like on upgrade new nodejs 10.x gets installed, so we're doing ok!

PR #242, opened 2018-11-28
Adds the universe repository to the used sources

owah:
This change is required, to support Ubuntu Server 18.04.01, which by default doesn't ship with universe. Universe contains python3-venv which is needed for tljh

### DIFF
--- a/bootstrap/bootstrap.py
+++ b/bootstrap/bootstrap.py
@@ -66,6 +66,7 @@ def main():
     else:
         logger.info('Setting up hub environment')
         initial_setup = True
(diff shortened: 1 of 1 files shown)


yuvipanda:
Thanks for the PR, @owah! It looks like add-apt-repository isn't available in the base ubunto docker image. I believe we have a function for adding repositories though, so maybe use that instead?

yuvipanda:
Picking this back up, I've rebased this on current master & installed the package that add-apt-repository needs. Let's see if this works!

This will conflict with #337 but should be an easy rebase.

PR #254, opened 2019-01-09
Install git explicitly during bootstrap

yuvipanda:
Some servers don't have git on by default


yuvipanda:
Unrelated test failures.

PR #262, opened 2019-01-29
Mark flaky test as flaky

yuvipanda:
We'll need to figure out why this is failing


minrk:
when conda's run as root, it respects the permissions in packages it extracts that are specified in the tarball. For some weird reason, some files in a few defaults channel packages have 777.

#268 is a more brute-force "Don't trust install permissions" that should catch similar issues that could come from conda or pip.

yuvipanda:
Thanks @minrk! I've merged #268 too!

PR #266, opened 2019-01-30
Replace chp with traefik-proxy

GeorgianaElena:
 - [x] Add tests

Hi!

I replaced configurable-http-proxy with a proxy that uses traefik and a toml file to store the routes (rules.toml), regarding https://github.com/jupyterhub/traefik-proxy/issues/32.
The proxy is available [here](https://github.com/jupyterhub/traefik-proxy).
Also, I managed to find a way to wait for the hub to be fully operational after a restart and got rid of the sleeps in the tests (the 1 sec sleep wasn't enough for traefik proxy)

```
assert 0 == await (await asyncio.create_subprocess_exec(*TLJH_CONFIG_PATH, 'reload')).wait()
# FIXME: wait for reload to finish & hub to come up
# Should be part of tljh-config reload
await asyncio.sleep(1)
```

In order to make sure the hub is ready after the restart, I first waited for jupyterhub service to be active and then I waited for the string "JupyterHub is now running at" to be available in jupyterhub's logs(since last restart time). @minrk, @yuvipanda, I'm not sure I chose the right approach for this, so a feedback would be highly appreciated.

Also, I don't understand why the test `test_hub.py/test_user_admin_remove` is failing. I noticed this test was also failing before this PR.
In the logs I see this:

```
[SingleUserNotebookApp auth:866] oauth state 'eyJ1dWlkIjogImZhYTAxZmRhODc4YTQ2ZjFhMGE0ZDMzMjE5Njc3YWExIiwgIm5leHRfdXJsIjogIi91c2VyLzhjYjk0ZTBlMzZkOTVjZjYvdHJlZSJ9' != None`
[SingleUserNotebookApp web:1667] 403 GET /user/8cb94e0e36d95cf6/oauth_callback?code=9ec9fca1-0b50-4b23-9692-ad1e222866d0&state=eyJ1dWlkIjogImZhYTAxZmRhODc4YTQ2ZjFhMGE0ZDMzMjE5Njc3YWExIiwgIm5leHRfdXJsIjogIi91c2VyLzhjYjk0ZTBlMzZkOTVjZjYvdHJlZSJ9 (127.0.0.1): oauth state does not match. Try logging in again.
```


tljh/systemd.py
```diff
@@ -70,3 +70,48 @@ def enable_service(name):
...
+
+        out = subprocess.check_output([
+            'journalctl',
+            '-u',
+            'jupyterhub',
+            '--since',
+            last_restart
+        ])
+
+        if "JupyterHub is now running at" in out.decode():
```
minrk: Rather than checking the logs, maybe check the HTTP endpoint with requests?

tljh/traefik.py
```diff
@@ -78,6 +78,9 @@ def ensure_traefik_config(state_dir):
         os.fchmod(f.fileno(), 0o744)
         f.write(new_toml)
```
minrk: I think we may need this one and below to have permissions `0o740` instead of `744` since there could be credentials in there now

minrk: or `0o600` to be safe

tljh/traefik.toml.tpl
```diff
@@ -33,14 +33,24 @@ idleTimeout = "10m0s"
...
   {% if https['tls']['cert'] %}
     [[entryPoints.https.tls.certificates]]
       certFile = "{{https['tls']['cert']}}"
       keyFile = "{{https['tls']['key']}}"
   {% endif %}
   {% endif %}
+  [entryPoints.auth_api]
+  address = ":8099"
+  [entryPoints.auth_api.auth.basic]
+  users = ["api_admin:$apr1$eS/j3kum$q/X2khsIEG/bBGsteP.x./"]
```
minrk: This credential will want to come from configuration, rather than hard-coded.

tljh/traefik.toml.tpl
```diff
@@ -57,13 +67,5 @@ entryPoint = "https"
...
-[frontends]
-  [frontends.jupyterhub]
-  backend = "jupyterhub"
-  passHostHeader = true
-[backends]
-  [backends.jupyterhub]
-    [backends.jupyterhub.servers.chp]
-    url = "http://127.0.0.1:15003"
-
+filename = "/opt/tljh/state/rules.toml"
```
minrk: I think this is going to need to take `TLJH_INSTALL_DIR` into account

tljh/traefik.toml.tpl
```diff
@@ -33,14 +33,24 @@ idleTimeout = "10m0s"
...
-  backend = "jupyterhub"
   [entryPoints.https.tls]
   {% if https['tls']['cert'] %}
     [[entryPoints.https.tls.certificates]]
       certFile = "{{https['tls']['cert']}}"
       keyFile = "{{https['tls']['key']}}"
   {% endif %}
   {% endif %}
+  [entryPoints.auth_api]
+  address = ":8099"
```
minrk: this port may want to come from config, too, and should probably include an ip that should default to 127.0.0.1 to avoid making the admin api available from outside by default.

tljh/config.py
```diff
@@ -183,11 +184,14 @@ def reload_component(component):
     from tljh import systemd, traefik
     if component == 'hub':
         systemd.restart_service('jupyterhub')
-        # FIXME: Verify hub is back up?
+        # Ensure hub is back up
+        while not systemd.check_service_active('jupyterhub'):
+            asyncio.sleep(1)
```
minrk: Since this is a blocking function not a coroutine, let's use `time.sleep` here instead of `asyncio.sleep`

tljh/traefik.toml.tpl
```diff
@@ -33,14 +33,26 @@ idleTimeout = "10m0s"
...
-  backend = "jupyterhub"
   [entryPoints.https.tls]
   {% if https['tls']['cert'] %}
     [[entryPoints.https.tls.certificates]]
       certFile = "{{https['tls']['cert']}}"
       keyFile = "{{https['tls']['key']}}"
   {% endif %}
   {% endif %}
+  [entryPoints.auth_api]
+  address = ":{{auth_api['port']}}"
```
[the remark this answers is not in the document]
GeorgianaElena: @minrk, I tried to achieve this with this config option for `auth_api`:
```
"whiteList": {"sourceRange": ["127.0.0.1"]}
```
Have you tested if defining the address of the entrypoint using `ip:port` works? I thought it doesn't :astonished:

yuvipanda:
@GeorgianaElena this is awesome work! I'm very happy to have it here :)

@minrk I think it's ok to switch to it unconditionally. Here's what I think we should do before that:

- [ ] Make sure upgrades work fine between CHP and this
- [ ] Run http://github.com/yuvipanda/hubtraf against this for a while with users coming and going to make sure it works fine. 

How does that sound?

minrk:
Sounds good. I'll run a couple of those tests.

tljh/traefik.toml.tpl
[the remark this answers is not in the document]
minrk: Yes, `ip:port` works. I'm not sure why they haven't documented that. `:443` is a standard way to express "this port on all ips" which is usually what you want for public services, but we know it should be private in this case.

tljh/config.py
```diff
@@ -172,6 +173,14 @@ def remove_config_value(config_path, key_path, value):
     with open(config_path, 'w') as f:
         yaml.dump(config, f)
 
+def check_hub_ready():
+    import requests
+
+    try:
+        r = requests.get('http://127.0.0.1:80')
```
minrk: to avoid failures when https doesn't work:

```suggestion
        r = requests.get('http://127.0.0.1:80', verify=False)
```

tljh/installer.py
```diff
@@ -133,28 +130,16 @@ def ensure_jupyterhub_service(prefix):
         jupyterhub_config_path=os.path.join(HERE, 'jupyterhub_config.py'),
         install_prefix=INSTALL_PREFIX,
     )
-    systemd.install_unit('configurable-http-proxy.service', proxy_unit_template.format(**unit_params))
```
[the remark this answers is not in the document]
GeorgianaElena: @minrk, I don't quite understand what you mean by leaving the configurable-http-proxy unit in place. Does that mean keep CHP as the default proxy and add an upgrade option somewhere in the installer that would stop CHP and use only traefik?
I've submitted a new commit addressing the other changes you suggested.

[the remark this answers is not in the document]
minrk: Ah, I mean that after upgrading from a version of tljh with CHP, to this PR with only traefik, configurable-http-proxy is still running from the previous version, even though it's not being used anymore. It would be good to clean up after ourselves. We should add a `systemd.uninstall_unit` that removes and unregisters the configurable-http-proxy unit while we are upgrading.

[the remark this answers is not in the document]
GeorgianaElena: I've noticed there's already an ```uninstall_unit``` in ```systemd``` that we could use.

Please correct me if I got this wrong: basically, what we want is to keep ```configurable-http-proxy.service``` and maybe add to it an ```ExecStop``` so that when uninstall_unit gets called, CHP will be properly stopped?
Also, is it ok to make use of the flag ```initial_setup``` in ```bootstrap.py``` to identify an upgrade and based on this call ```systemd.uninstall_unit```?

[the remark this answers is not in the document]
yuvipanda: Ideally, I'd love for us to not have to differentiate between initial_setup and upgrades wherever possible. In the installer, the general way to do that is to just uninstall the unit if it exists, regardless of initial setup or upgrade. This keeps the install / upgrade process idempotent.

tljh/installer.py
```diff
@@ -99,15 +99,25 @@ def ensure_node():
...
+def remove_chp():
     """
-    Ensure CHP is installed
+    Ensure CHP is not running
     """
-    if not os.path.exists(os.path.join(prefix, 'node_modules', '.bin', 'configurable-http-proxy')):
-        subprocess.check_output([
-            'npm', 'install', 'configurable-http-proxy@3.1.0'
-        ], cwd=prefix, stderr=subprocess.STDOUT)
+    if os.path.exists(os.path.join(HERE, 'systemd-units', 'configurable-http-proxy.service')):
```
minrk: This is not the path where the unit is installed, which is what we want to check:

```suggestion
    if os.path.exists("/etc/systemd/system/configurable-http-proxy.service"):
```

minrk:
@GeorgianaElena this is great! I think it's ready to go. I added a small change that fixed the issue, which was that the jupyterhub_config.py wasn't loading the generated password. I also moved the generation of the auth_basic bit to traefik.py where we generate that config file, so that secrets get loaded as part of the 'user' configuration instead of the 'default' configuration, which removed the need for mocking out the `generate_traefik_api_credentials`

GeorgianaElena:
Thanks a lot for your help, @minrk!

minrk:
Running hubtraf with jupyterhub-traefik-proxy 0.1.2 (needs https://github.com/jupyterhub/traefik-proxy/pull/51) resulted in 0 failures for 20, 40, and 60 users, so I think we are set here.

minrk:
Great job, @GeorgianaElena!

yuvipanda:
This is *awesome*, @GeorgianaElena / @minrk! How would you feel about writing a blog post about this work?

GeorgianaElena:
Thank you @minrk, @yuvipanda!
@yuvipanda , I would really like to give it a try.

yuvipanda:
@GeorgianaElena awesome. If you can make an account on medium.com and tell us, we can give you access to send stories to blog.jupyter.org. Then you can write it, a bunch of us can help edit it and publish it. How does that sound?

/cc @choldgraf who is amazing and helps with these things.

choldgraf:
definitely more than happy to help writing/editing/whatever's useful!

minrk:
@GeorgianaElena's draft is submitted to the Jupyter blog if folks want to give it a review. I think we can post it early next week.

PR #276, opened 2019-02-12
update dev-setup commands

minrk:
- Dockerfile moved to integration-tests
- remove sudo from commands (usually not needed, better added by users where it is)


leportella:
Tested and this solution works :)

PR #283, opened 2019-02-20
[WIP] Change to native authenticator

leportella:
 - [X] Add tests

### DIFF
--- a/tests/test_configurer.py
+++ b/tests/test_configurer.py
@@ -106,9 +106,8 @@ def test_auth_default():
     """
     c = apply_mock_config({})
 
-    assert c.JupyterHub.authenticator_class == 'firstuseauthenticator.FirstUseAuthenticator'
-    # Do not auto create users who haven't been manually added by default
-    assert not c.FirstUseAuthenticator.create_users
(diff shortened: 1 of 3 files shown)


tests/test_configurer.py
```diff
@@ -127,20 +126,20 @@ def test_auth_dummy():
     assert c.DummyAuthenticator.password == 'test'
 
 
-def test_auth_firstuse():
+def test_auth_nativeauth():
```
yuvipanda: I think this test should be in addition to the firstuse test, not replace it. We'll stil ship first use authenticator.

tljh/configurer.py
```diff
@@ -17,10 +17,8 @@
 # User provided config is merged into this
 default = {
     'auth': {
-        'type': 'firstuseauthenticator.FirstUseAuthenticator',
-        'FirstUseAuthenticator': {
-            'create_users': False
-        }
+        'type': 'nativeauthenticator.NativeAuthenticator',
```
yuvipanda: I don't think we should switch the default yet! This PR can be just adding it as an option, and adding docs. We can then work on migration pathway, and then make it default.

leportella:
@yuvipanda since I have to change both commits, I'll create a new pull request ok!?

Issue #264, opened 2019-01-29
Switch to using NativeAuthenticator as default


yuvipanda:
@leportella  has been working on the awesome http://github.com/jupyterhub/nativeauthenticator/. We should switch to using that as the default authenticator here rather than FirstUseAuthenticator.

Tasks
- [ ] Install nativeauthenticator by default
- [ ] Try find a way to seamlessly import FirstUseAuthenticator credentials into NativeAuthenticator - makes upgrades seamless for users.
- [ ] Write docs on how NativeAuthenticator works
- [ ] Switch to nativeauthenticator by default

mtav:
- [ ] Expand the documentation for nativeauthenticator a bit, because it is our default authenticator. 

Maybe this could even be done before switching to NativeAuthenticator by default. I certainly could have used it this week!
See [here](https://github.com/jupyterhub/the-littlest-jupyterhub/issues/295#issuecomment-1242281880) for more details.

I would suggest not only documenting specifically NativeAuthenticator configuration, but also how to properly set lists with tljh-config (in doc or in _tljh-config -h_), i.e. properly go from python code in other authenticator's documentations to tljh-config commands.
Or just directly document how to do it _jupyterhub_config.d_ ?

PR #284, opened 2019-02-20
Add NativeAuth as an optional authenticator

leportella:
- [X] Add / update documentation
 - [X] Add tests

It closes 2 tasks from #264

### DIFF
--- a/docs/howto/auth/nativeauth.rst
+++ b/docs/howto/auth/nativeauth.rst
@@ -1,8 +1,8 @@
 .. _howto/auth/nativeauth:
 
-==================================================
-Users can signup and create username and password
-==================================================
+==============================================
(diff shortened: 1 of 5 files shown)


docs/howto/auth/nativeauth.rst
```diff
@@ -0,0 +1,38 @@
+.. _howto/auth/nativeauth:
+
+==================================================
+Users can signup and create username and password
```
yuvipanda: What do you think of making the title be 'let users sign up with a username and password'? This way, it completes the sentence 'how do I <title>', which is how other authenticator docs are titled.

leportella: Just changed it :)

PR #287, opened 2019-03-05
typo fix in installer actions

junctionapps:
minor typo fix


betatim:
Restarted the CI as this change shouldn't have lead to a failure.

betatim:
@junctionapps not sure why this fails. Maybe rebasing this on the current master will "fix it" (aka have you tried restarting it?). Otherwise someone will have to dive a bit deeper into the integration tests to figure out why they are failing, it is definitely not related to the change in this PR but something "external".

Issue #290, opened 2019-03-07
Can't connect to the kernel


PaluchowskiMatthew:
I successfully installed jupyterhub on a google cloud instance however I am unable to start a working jupyter notebook as it fails to connect to a kernel with a following message:

> Connection failed A connection to the notebook server could not be established. The notebook will continue trying to reconnect. Check your network connection or notebook server configuration.

![image](https://user-images.githubusercontent.com/13195985/53955593-b5948080-40d9-11e9-8654-1afcf2bd26d3.png)

 jupyterhub.service logs: 
> root@jupyterhub-0:/opt/tljh/config/jupyterhub_config.d# sudo journalctl -u jupyterhub.service
-- Logs begin at Thu 2019-03-07 11:12:13 UTC, end at Thu 2019-03-07 12:07:43 UTC. --
Mar 07 11:16:46 jupyterhub-0 systemd[1]: /etc/systemd/system/jupyterhub.service:12: Failed to parse protect home value, ignoring: tmpfs
Mar 07 11:16:46 jupyterhub-0 systemd[1]: Started jupyterhub.service.
Mar 07 11:16:46 jupyterhub-0 systemd[1]: Stopping jupyterhub.service...
Mar 07 11:16:46 jupyterhub-0 systemd[1]: Stopped jupyterhub.service.
Mar 07 11:16:46 jupyterhub-0 systemd[1]: Started jupyterhub.service.
Mar 07 11:16:46 jupyterhub-0 systemd[1]: /etc/systemd/system/jupyterhub.service:12: Failed to parse protect home value, ignoring: tmpfs
Mar 07 11:16:47 jupyterhub-0 systemd[1]: /etc/systemd/system/jupyterhub.service:12: Failed to parse protect home value, ignoring: tmpfs
Mar 07 11:16:48 jupyterhub-0 python3[7815]: [I 2019-03-07 11:16:48.206 JupyterHub app:1673] Using Authenticator: firstuseauthenticator.firstuseauthenticator.FirstUseAuthenticator
Mar 07 11:16:48 jupyterhub-0 python3[7815]: [I 2019-03-07 11:16:48.207 JupyterHub app:1673] Using Spawner: builtins.UserCreatingSpawner
Mar 07 11:16:48 jupyterhub-0 python3[7815]: [I 2019-03-07 11:16:48.209 JupyterHub app:1055] Writing cookie_secret to /opt/tljh/state/jupyterhub_cookie_secret
[...]

PR #292, opened 2019-03-08
Pin tornado to <6

willirath:
See #290.  Users reported Kernel connection issues with tornado >= v6.  Let's see if this fixes it.


PaluchowskiMatthew:
@willirath 
problem persists and tornado version remains 6.0.1
image

willirath:
@PaluchowskiMatthew thanks for checking.  With af26b22, tornado should be pinned in the user env.

yuvipanda:
Thank you, @willirath

PR #294, opened 2019-03-14
DOC: moved nativeauthentic config instructions to code block

story645:
Moved the config to a code block because as rendered, the commands blend and create an error when trying to c&p. Also removed the 1. because there's no 2. Here's as currently rendered in docs:
image

And here's the change:
image

 - [ x ] Add / update documentation
 - [ ] Add tests


yuvipanda:
Welcome to the littlest jupyterhub, @story645! 

Thank you very much for the PR. Look forward to more PRs :D

leportella:
Thanks  @story645 !!

PR #303, opened 2019-04-10
[MRG] Fix the version tag of the notebook package

betatim:
This takes TLJH to notebook 5.7.8. According to PyPI 5.7.7 was never released so I think this fixes typo.


willirath:
Maybe 5.7.7 was deleted for security reasons?

willirath:
Thanks! Merging.

betatim:
I thought once published on PyPI you couldn't pull a release again. Maybe you can pull it but not overwrite it :-/

PR #309, opened 2019-04-26
adding sidebar links

choldgraf:

This does a couple things:

1. Adds sidebar links for the table of contents, so you can navigate to other pages from within sub-pages
2. Adds captions to the toctrees, so that there are headers in the sidebar links
3. Adds css to make the within-page captions invisible because we're already using headers for this instead of captions

curious what @yuvipanda or @leportella think about this!


yuvipanda:
I like it

PR #313, opened 2019-05-03
Upgrade to JupyterHub 1.0

minrk:
bump oauthenticator patch as well

### DIFF
--- a/tljh/installer.py
+++ b/tljh/installer.py
@@ -189,13 +189,13 @@ def ensure_jupyterhub_package(prefix):
     and conda packages!
     """
     conda.ensure_pip_packages(prefix, [
-        'jupyterhub==0.9.6',
+        'jupyterhub==1.0.0',
         'jupyterhub-dummyauthenticator==0.3.1',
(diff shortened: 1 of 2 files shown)


yuvipanda:
@minrk awesome! Do you think we should put upgrade-db in pre-start for JupyterHub service here, similar to what we do in z2jh?

minrk:
Yeah, that's a good idea.

yuvipanda:
@minrk let me know if my change looks good

yuvipanda:
Not sure yet why this is failing.

minrk:
Debugging the integration tests can be hard, since we get little feedback about the errors. It would be good to show logs on failure. I'll look into it today.

yuvipanda:
@minrk did you manage to find any time to look into this?

yuvipanda:
Instead of using pre-start, I have instead passed --upgrade-db to the JupyterHub process itself. This matches what we do in z2jh.

yuvipanda:
Tests pass, yay!

yuvipanda:
I'll merge this one now :)

Issue #257, opened 2019-01-17
Installation fails due to not enough memory on host


hoeller:
Hi!
I have problems installing on a fresh Ubuntu Server 18.04 LTS VM following [this tutorial.](https://tljh.jupyter.org/en/latest/install/custom-server.html#step-1-installing-the-littlest-jupyterhub)
Seems that the `conda install` gets aborted?
Output is:
```
Checking if TLJH is already installed...
TLJH already installed, upgrading...
Upgrading TLJH installer...
Setup tljh package
Starting TLJH installer...
Setting up admin users
Granting passwordless sudo to JupyterHub admins...
Setting up user environment...
Traceback (most recent call last):
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/opt/tljh/hub/lib/python3.6/site-packages/tljh/installer.py", line 448, in <module>
    main()
  File "/opt/tljh/hub/lib/python3.6/site-packages/tljh/installer.py", line 430, in main
    ensure_user_environment(args.user_requirements_txt_url)
  File "/opt/tljh/hub/lib/python3.6/site-packages/tljh/installer.py", line 234, in ensure_user_environment
    'conda==4.5.8'
  File "/opt/tljh/hub/lib/python3.6/site-packages/tljh/conda.py", line 95, in ensure_conda_packages
    ] + packages).decode()
  File "/usr/lib/python3.6/subprocess.py", line 336, in check_output
    **kwargs).stdout
  File "/usr/lib/python3.6/subprocess.py", line 418, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['/opt/tljh/user/bin/python', '-m', 'conda', 'install', '-c', 'conda-forge', '--json', '--prefix', '/opt/tljh/user', 'conda==4.5.8']' died with <Signals.SIGKILL: 9>.
```

yuvipanda:
To resurrect this thread, I usually test on a VM with 512MB of RAM and it works ok... However, it's possible that conda's memory usage has ballooned since. I'm re-opening this to make sure we can test how much RAM is needed, and document it in the appropriate place.

Thank you very much for filing the issue, and then posting the solution that worked for you!

yuvipanda:
[...]

PR #336, opened 2019-05-19
[WIP] Bump version of conda

yuvipanda:
We're on a pretty old version. 4.6 brings with it a host
of performance improvements, and hopefully memory improvements
(for #257)

 - [ ] Add tests

### DIFF
--- a/tljh/installer.py
+++ b/tljh/installer.py
@@ -222,8 +222,8 @@ def ensure_user_environment(user_requirements_txt_file):
     Set up user conda environment with required packages
     """
     logger.info("Setting up user environment...")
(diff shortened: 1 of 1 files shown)


yuvipanda:
Need to check if this also bumps version of Python.

yuvipanda:
It does bump the version of Python. I am not sure what exactly to do here. I'm guessing if you're already installed libraries and what not with a 3.6 version of python in your user directory, having that be upgraded automatically to 3.7 is not a good idea... Maybe we need separate 'conda' and 'user' directories here like what @minrk did for repo2docker.

PR #337, opened 2019-05-19
Provide much better error messages

### LINKED ISSUE #84: Write stdout / stderr from all subprocess commands to log
In bootstrap / installer, we call a lot of external commands. However, we do *not* log their output to the logs - this means we lose very important information there. We should be logging *all* that too - but not outputting it to stdout.

### LINKED ISSUE #142: installer.log should not be world readable
Since the the `/opt/tljh/` hosted configuration files (at least `/opt/tljh/` and `/opt/tljh/jupyterhub_config.d` can contain information on secret keys for authentication and other sensitive information, the file permissions for those configuration files should be tightened down so that non-admin users can't read those files if possible.  

My own testing shows there doesn't seem to be a problem with restricting reading to `root`.

yuvipanda:
- When processes fail, they actually print a failure message
  on the user's terminal
- Regardless of success or failure, we print all output to
  /opt/tljh/installer.log
- Make sure installer.log isn't readable by non-root users

This should make debugging people's issues *much* easier, since
we can actually see the output of failing commands rather than
having to guess.

Fixes #84 
Fixes #142

 - [x] Add tests

### DIFF
--- a/bootstrap/bootstrap.py
+++ b/bootstrap/bootstrap.py
@@ -18,6 +18,7 @@ import sys
 import logging
 import shutil
 
+logger = logging.getLogger(__name__)
 
 def get_os_release_variable(key):
     """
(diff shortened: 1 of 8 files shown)


yuvipanda:
Rebased.

yuvipanda:
<3 thanks, @betatim

Issue #16, opened 2018-07-02
Better error handling in bootstrap script


yuvipanda:
Provide nicer error messages, and check for error conditions properly in the bootstrap script.

yuvipanda:
Things to check:

1. Ubuntu
2. Appropriate version of Ubuntu
3. Git installed
4. Appropriate version of Python
5. Systemd installed and present (accounts for running inside docker container)

yuvipanda:
From https://github.com/jupyterhub/the-littlest-jupyterhub/issues/219#issuecomment-446774526, we should also check for external internet access & fail if we can't reach it. (Thanks @fm75)

yuvipanda:
We already do the following:

1. Validate it's ubuntu
2. Validate version of ubuntu
3. We install git ourselves, so that's ok

yuvipanda:
I'm not particularly sure how to check for 'do we have internet access'. I don't want to do something like 'ping google.com' or any particular external website, since that will leak information about the server being installed. I guess we should watch for apt failures and suggest that as a possible solution.

yuvipanda:
#340 adds validation for python version and systemd presence. It also handles erroring with a more useful error message when running inside docker.

Issue #333, opened 2019-05-19
Make it clearer we don't support running TLJH *inside* a docker container


yuvipanda:
TLJH depends on Systemd, and hence can't run inside a Docker container without a bunch of extra hacks. It'll fail in mysterious ways if you try. We should make it clearer in the docs, and maybe add a flag to bootstrap.py as well.

GeorgianaElena:
Fixed by https://github.com/jupyterhub/the-littlest-jupyterhub/pull/327

PR #340, opened 2019-05-20
Add more validation to bootstrap.py

yuvipanda:
- Check for Python Version
- Check if systemd is present
- Provide more useful error message when running inside
  an unprepared docker container

Ref #16

 - [x] Add tests

### DIFF
--- a/bootstrap/bootstrap.py
+++ b/bootstrap/bootstrap.py
@@ -16,6 +16,7 @@ import os
 import subprocess
 import sys
 import logging
+import shutil
 
 
 def get_os_release_variable(key):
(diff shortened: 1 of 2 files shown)


yuvipanda:
Docs on docker container in #333. This still needs tests though.

yuvipanda:
Test added. I'm curious about the wording. I'm not sure 'plain' docker container is quite right.

betatim:
Not quite sure what it means to run in "plain docker". It makes me think that this is using a version of docker that hasn't been modified. As in the "plain" refers to dockerd, not the image that we are using (which is what I think this is about).

Could it be named "no-systemd-docker"?

yuvipanda:
@betatim updated to refer to 'docker container without systemd' instead.

Issue #203, opened 2018-10-15
How to unset a config value


betatim:
The [configuration documentation](https://the-littlest-jupyterhub.readthedocs.io/en/latest/topic/tljh-config.html) does not explain how to unset/remove a value after setting it.

PR #352, opened 2019-05-28
Add "tljh-config unset" option

GeorgianaElena:
 - [x] Add tests

closes https://github.com/jupyterhub/the-littlest-jupyterhub/issues/203

I added the option to unset (remove) a config property. @yuvipanda, I'm not sure if this is wanted but I thought it would be nice to be able to remove a config property the same way we're setting it, without editing the config file.

### DIFF
--- a/docs/topic/tljh-config.rst
+++ b/docs/topic/tljh-config.rst
@@ -22,8 +22,8 @@ You can run ``tljh-config`` in two ways:
 .. _tljh-set:
 
 
-Set a configuration property
-============================
+Set / Unset a configuration property
+====================================
 
 TLJH's configuration is organized in a nested tree structure. You can
 set a particular property with the following command:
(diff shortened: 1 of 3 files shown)


yuvipanda:
This is very useful! Thank you very much :) <3 

We should probably also have a way to add an explicit key-value pair to config as well. Currently we can't set [LTIAuthenticator](https://github.com/jupyterhub/ltiauthenticator) properties, since they look like:

```python
c.LTIAuthenticator.consumers = {
	'key1': 'value1',
	'key2': 'value2'
}
```

Not sure what the terminal UX for this would look like.

GeorgianaElena:
@yuvipanda, I'm not sure I understand why the following won't do the job.
```
tljh-config set c.LTIAuthenticator.consumers.key1 = value1
tljh-config set c.LTIAuthenticator.consumers.key2 = value2
``` 
Is it because we assume the authenticator will will only have one-level attributes?

When you say
>add an explicit key-value pair to config 

do you mean adding a key as it is without parsing it? If this is the case we could use the set command and add a flag that would tell us whether or not to treat the key as a path.

Sorry if I'm far from the truth and you meant something else.

PR #360, opened 2019-05-31
Add custom hub package & config hooks

yuvipanda:
- Add hook for installing additional hub packages
- Add hook for additional `jupyterhub_config.py`
  entries

 - [x] Add tests

### DIFF
--- a/integration-tests/test_simplest_plugin.py
+++ b/integration-tests/test_simplest_plugin.py
@@ -2,6 +2,7 @@
 Test simplest plugin
 """
 from ruamel.yaml import YAML
+import requests
 import os
 import subprocess
 from tljh.config import CONFIG_FILE, USER_ENV_PREFIX, HUB_ENV_PREFIX
@@ -52,3 +53,11 @@ def test_config_hook():
         data = yaml.load(f)
 
     assert data['simplest_plugin']['present']
+
+def test_jupyterhub_config_hook():
+    """
+    Test that tmpauthenticator is enabled by our custom config plugin
+    """
+    resp = requests.get('http://localhost/hub/tmplogin', allow_redirects=False)
+    assert resp.status_code == 302
+    assert resp.headers['Location'] == '/hub/spawn'
--- a/integration-tests/test_simplest_plugin.py
+++ b/integration-tests/test_simplest_plugin.py
@@ -4,7 +4,7 @@ Test simplest plugin
 from ruamel.yaml import YAML
 import os
 import subprocess
-from tljh.config import CONFIG_FILE, USER_ENV_PREFIX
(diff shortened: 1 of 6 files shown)


yuvipanda:
Adding these hooks to help set up https://discourse.jupyter.org/t/would-a-the-littlest-binder-be-useful/1041/28.

integration-tests/test_simplest_plugin.py
```diff
@@ -46,3 +53,11 @@ def test_config_hook():
         data = yaml.load(f)
 
     assert data['simplest_plugin']['present']
+
+def test_tmpauthenticator():
+    """
```
GeorgianaElena: I would probably name this *test_tmpauthenticator_enabled* or *test_jupyterhub_config_hook* (I'm not very good at naming things though :eyes: ) but the name as it is makes me think we're testing the authenticator.

tljh/installer.py
```diff
@@ -208,6 +208,7 @@ def ensure_jupyterhub_package(prefix):
         'jupyterhub-firstuseauthenticator==0.12',
         'jupyterhub-nativeauthenticator==0.0.4',
         'jupyterhub-ldapauthenticator==1.2.2',
+        'jupyterhub-tmpauthenticator==0.6',
```
GeorgianaElena: From what I understand, when a user wants to custom its juputerhub_config using *tljh_custom_jupyterhub_config* in order to use something that's not directly installed by TLJH, they could would use *tljh_extra_hub_pip_packages* to install it first. 

If this is the case, I believe tmpauthenticator package should/could come from *tljh_extra_hub_pip_packages*.

integration-tests/test_simplest_plugin.py
yuvipanda: Good catch! I've renamed it!

tljh/installer.py
yuvipanda: I've split this part of the PR to https://github.com/jupyterhub/the-littlest-jupyterhub/pull/365 separately, and based this PR on that. So we have two PRs:

1. #365 that adds tmpauthenticator
2. This PR, which adds two hooks with tests for both of them. There is a test for `tljh_extra_hub_pip_packages` that imports the 'there' package, and the test for `tljh_custom_jupyterhub_config` that sets tmpauthenticator to be the default authenticator.

How does that sound?

yuvipanda:
Thanks for the review, @GeorgianaElena! I've updated the PR to reflect your suggestions. Let me know what you think :)

#365 should probably be merged before this though - this PR 'depends' on that (it includes some commits from there). So once #365 gets merged, the diff on this PR should stop showing the contents of #365. This might be confusing - let me know if I can help clarify that.

GeorgianaElena:
Excellent @yuvipanda :confetti_ball:. Once the tests pass I'll try to merge the PRs in the order you suggested :D

PR #364, opened 2019-06-05
Docs addition

kafonek:
Update to the docs for plugin information and "shared directory" configurations.  

Apologies if the `.rst` format is slightly off, it was hard to evaluate what it will look like in tljh.jupyter.org since there's no jupyterlab renderer and github rendering of rst is a bit different.


GeorgianaElena (approved):
Hey @kafonek :wave: Sorry for the late reply.
I've rebased this PR and added a few commits to make it match the current docs structure. Hope that's ok.
Merging it now. Thanks a lot :tada:

PR #366, opened 2019-06-07
Add idle culler

GeorgianaElena:
@yuvipanda, I decided to give this a try :D
However, I have some questions and doubts.
Is there a better way to access the *cull_idle_servers.py*, without actually copying it to TLJ?
I'm not sure about the test either... Is it ok for the it to be part of *test_hub.py* since the idle culler is managed by JupyterHub?  Also, I'm not sure I'm testing correctly that the user's server was culled.
 
 - [x] Add tests

### DIFF
--- a/docs/topic/idle-culler.rst
+++ b/docs/topic/idle-culler.rst
@@ -1,0 +1,86 @@
+.. _topic/idle-culler:
+
+=============================
+Culling idle notebook servers
+=============================
+
+The idle culler is a hub-managed service that automatically shuts down idle
+single-user notebook servers in order to free up resources. After culling, any
+in-memory data will be lost.
+
+
+Disabling the idle culler
+=========================
+
+The idle culling service is enabled by default. To disable it, use the following
+command:
+
+.. code-block:: bash
+
+   sudo tljh-config set services.cull.enabled False
+
+
+Configuring the idle culler
+===========================
+
+By **default**,  JupyterHub will:
+	* Run the culling process every minute.
+	* Cull any user servers that have been inactive for more than 10 minutes.
+
+The configuration options available are:
+
+Idle timeout
+------------
+
+The idle timeout (in seconds) can be configured using:
+
+.. code-block:: bash
+
+	sudo tljh-config set services.cull.timeout <max-idle-sec-before-server-is-culled>
+
+*By default services.cull.timeout = 600*
+
+Idle check interval
+-------------------
+
+The interval (in seconds) for checking for idle servers to cull can be configured using:
+
+.. code-block:: bash
+
+ 	sudo tljh-config set services.cull.every <number-of-sec-this-check-is-done>
+
+*By default services.cull.every = 60*
+
+Maximum age
+-----------
+
+The maximum age (in seconds) of servers that should be culled even if they are active
+can be configured using:
+
+.. code-block:: bash
+
+ 	sudo tljh-config set services.cull.max_age <server-max-age>
+
+*By default services.cull.max_age = 0*
+
+User culling
+------------
+
+In addition to servers, the users will also be culled if the following command is used:
+
+.. code-block:: bash
+
+ 	sudo tljh-config set services.cull.users True
+
+*By default services.cull.users = False*
+
+Concurrency
+-----------
+
+The number of concurrent requests made to the Hub ca be configured using:
+
+.. code-block:: bash
+
+ 	sudo tljh-config set services.cull.concurrency <number-of-concurrent-hub-requests>
+
+*By default services.cull.concurrency = 5*
--- a/docs/topic/idle-culler.rst
+++ b/docs/topic/idle-culler.rst
@@ -4,9 +4,102 @@
 Culling idle notebook servers
 =============================
 
-The idle culler is a hub-managed service that automatically shuts down idle
-single-user notebook servers in order to free up resources. After culling, any
-in-memory data will be lost.
+The idle culler automatically shuts down user notebook servers when they have
+not been used for a certain time period, in order to reduce the total resource
+usage on your JupyterHub.
+
+
+JupyterHub pings the user's notebook server at certain time intervals. If no response
+is received from the server during this checks and the timeout expires, the server is
+considered to be *inactive (idle)* and will be culled.
+
+
+Default settings
+================
+
+By default, JupyterHub will ping the user notebook servers every 60s to check their
+status. Every server found to be idle for more than 10 minutes will be culled.
+
+.. code-block:: python
+
+	services.cull.every = 60
+	services.cull.timeout = 600
+
+Because the servers don't have a maximum age set, an active server will not be shut down
+regardless of how long it has been up and running.
+
+.. code-block:: python
+
+	services.cull.max_age = 0
+
+If after the culling process, there are users with no active notebook servers, by default,
+the users will not be culled alongside their notebooks and will continue to exist.
+
+.. code-block:: python
+
+	services.cull.users = False
+
+
+Configuring the idle culler
+===========================
+
+The available configuration options are:
+
+Idle timeout
+------------
+The idle timeout is the maximum time (in seconds) a server can be inactive before it
+will be culled. The timeout can be configured using:
+
+.. code-block:: bash
+
+	sudo tljh-config set services.cull.timeout <max-idle-sec-before-server-is-culled>
+	sudo tljh-config reload
+
+Idle check interval
+-------------------
+The idle check interval represents how frequent (in seconds) the Hub will
+check if there are any idle servers to cull. It can be configured using:
+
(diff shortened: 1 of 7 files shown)


tljh/configurer.py
```diff
@@ -191,6 +203,43 @@ def update_traefik_api(c, config):
...
 
 
+def set_cull_idle_service(c, config):
+    """
+    Set Idle Culler service
+    """
+    cull_cmd = [
+       sys.executable, '/srv/src/tljh/cull_idle_servers.py'
+    ]
+    if config['services']['cull']['timeout']:
```
yuvipanda: Since this was set in the default, I *think* these properties will always be present even if the user doesn't override them, so you don't need all the conditionals.

yuvipanda: Since you are accessing `config['services']['cull']` a lot, it would make this more readable to pull it out into its own variable.

```python
cull_config = config['services']['cull']
```

tljh/configurer.py
```diff
@@ -191,6 +203,43 @@ def update_traefik_api(c, config):
...
+        cull_cmd.append('--timeout=%s' % config['services']['cull']['timeout'])
```
yuvipanda: I always try to avoid string interpolation in creatin commands, to protect against [command injection](https://www.owasp.org/index.php/Command_Injection). I *think* in this case we are safe, but would be better to do it like:

```python
cull_cmd += ['--timeout', str(cull_config['timeout']]
```

integration-tests/test_hub.py
```diff
@@ -137,4 +138,41 @@ def test_hub_up():
...
+            # Start user's server
+            await u.ensure_server()
+            # Assert that the user exists
+            assert pwd.getpwnam(f'jupyter-{username}') is not None
+
+            # Check that we can get to the user's server
+            r = await u.session.get(u.hub_url / 'hub/api/users' / username,
+                headers={'Referer': str(u.hub_url / 'hub/')})
+            assert r.status == 200
+            time.sleep(60)
```
yuvipanda: Since we're in an async function, `await asyncio.sleep(60)` is probably more appropriate.

tljh/configurer.py
GeorgianaElena: Thanks a lot for your feedback @yuvipanda :heart: 
One small question: are the default values ok? Should I change anything there?

integration-tests/test_hub.py
GeorgianaElena: @yuvipanda, I just noticed this test is failing intermittently :disappointed:
Do you think it would be wiser to use [exponential_backoff](https://github.com/jupyterhub/jupyterhub/blob/af0082a16ba365a4b1bfb79f9d01a7c1a6cee5be/jupyterhub/utils.py#L93) here with maybe an 100s timeout, instead of relying on the 60s sleep to be enough for the server to be culled?

yuvipanda: Ah yeah, that makes sense! I think we should have two tests:

1. One with a 100s timeout to make sure the server is culled
2. Another, with a 30s timeout, to make sure the server is *not* culled

I think this will give us a lot more confidence in the code. what do you think?

tljh/configurer.py
yuvipanda: yw! Thanks for working on this :) the defaults look ok to me as long as we document them.

integration-tests/test_hub.py
GeorgianaElena: This sounds perfect :D

tljh/configurer.py
GeorgianaElena: @yuvipanda, I noticed that [parse_command_line](https://www.tornadoweb.org/en/stable/_modules/tornado/options.html#OptionParser.parse_command_line) used by *cull_idle_servers.py* requires the options to be passed in like ``--option=value`` and complains about having them like ``--option value``.
I could enforce the type of the option value in *cull_idle_servers.py* like this:
```
define('timeout', type=int, default=600, help="The idle timeout (in seconds)")
```
and then add it to the command like this:
```
cull_cmd += ['--timeout=%d' % cull_config['timeout']]
```
Do you believe this approach will reduce enough the risks of a command injection ?

GeorgianaElena: @yuvipanda, I've addressed your feedback and defined the *cull_cmd* as described above. Please let me know what you think :D
About the documentation, where do you believe it is better to add it? I'm thinking about *Administration and security* or *Topic Guides* :thinking:

yuvipanda: @GeorgianaElena great! I'll take a look shortly. I think 'Topic' is a great place to start here.

docs/topic/idle-culler.rst
```diff
@@ -0,0 +1,86 @@
+.. _topic/idle-culler:
+
+=============================
+Culling idle notebook servers
+=============================
+
+The idle culler is a hub-managed service that automatically shuts down idle
```
yuvipanda: We should split this into a couple parts, written as paragraphs / sections.


1. What idle culling does & why. Something like:

```
The idle culler automatically shuts down user notebook servers when they have
not been used for a certain time period to reduce total resource usage on
your JupyterHub. 
```

2. What constitutes 'idle'? When is that triggered? This should be a small section of its own.

3. What are the default settings? Should also be a section. It should be descriptive rather than just listing the settings themselves.

yuvipanda: We can iterate on these as we go forward and I'm totally happy to help write them too if you wish :)

docs/topic/idle-culler.rst
```diff
@@ -0,0 +1,86 @@
...
+
+.. code-block:: bash
+
+   sudo tljh-config set services.cull.enabled False
+
+
+Configuring the idle culler
+===========================
+
+By **default**,  JupyterHub will:
```
yuvipanda: See comment above on a specific 'defaults' section

docs/topic/idle-culler.rst
```diff
@@ -0,0 +1,86 @@
...
+	* Run the culling process every minute.
+	* Cull any user servers that have been inactive for more than 10 minutes.
+
+The configuration options available are:
+
+Idle timeout
+------------
+
+The idle timeout (in seconds) can be configured using:
```
yuvipanda: A quick description of what the idle timeout is here would be great. Repetition is fine for cases like this I think.

docs/topic/idle-culler.rst
```diff
@@ -0,0 +1,86 @@
...
+
+.. code-block:: bash
+
+	sudo tljh-config set services.cull.timeout <max-idle-sec-before-server-is-culled>
+
+*By default services.cull.timeout = 600*
```
yuvipanda: I don't think this needs to be in italics

docs/topic/idle-culler.rst
```diff
@@ -0,0 +1,86 @@
...
+.. code-block:: bash
+
+ 	sudo tljh-config set services.cull.every <number-of-sec-this-check-is-done>
+
+*By default services.cull.every = 60*
+
+Maximum age
+-----------
+
+The maximum age (in seconds) of servers that should be culled even if they are active
```
yuvipanda: Maybe add a line here on why you would do this?

docs/topic/idle-culler.rst
```diff
@@ -0,0 +1,86 @@
...
+.. code-block:: bash
+
+ 	sudo tljh-config set services.cull.max_age <server-max-age>
+
+*By default services.cull.max_age = 0*
+
+User culling
+------------
+
+In addition to servers, the users will also be culled if the following command is used:
```
yuvipanda: Add a line on why you would od this?

docs/topic/idle-culler.rst
```diff
@@ -0,0 +1,86 @@
...
+.. code-block:: bash
+
+ 	sudo tljh-config set services.cull.users True
+
+*By default services.cull.users = False*
+
+Concurrency
+-----------
+
+The number of concurrent requests made to the Hub ca be configured using:
```
yuvipanda: Also add a line on why you would do this? I think in the TLJH case this should be very rare...

docs/topic/idle-culler.rst
GeorgianaElena: Thanks a lot for your feedback @yuvipanda :heart:
I tried my best to address you suggestions, but I have a feeling the docs might also need your intervention :eyes:

yuvipanda: @GeorgianaElena This is perfect! I'm going to merge this now.

yuvipanda:
Congratulations on your second big PR being merged into TLJH, @GeorgianaElena! <3 \o/

gsemet:
Look great !

GeorgianaElena:
Yay <3 thanks a lot @yuvipanda

PR #381, opened 2019-06-18
Test TLJH against 19.04

yuvipanda:
It might be useful to test against latest LTS
as well as latest Ubuntu. This makes life easier
when we wanna update to new LTS, and provides
an option for people who want more recent package
versions

 - [ ] Add tests

TODO:
- [ ] Don't stop testing 18.04
- [ ] Document that you can run this on 19.04, and why you would


minrk:
From the outputs, looks like `wheel` is a missing dependency?

jtpio:
Out of curiosity I tried it on an OVH b2-7 Ubuntu 19.04 and the install went through.

Issue #349, opened 2019-05-24
add jupyter-* to some group


gsemet:
Can tljh support automatically add its `jupyer-*` users to some groups (like "video" or "cuda", for instance)?

yuvipanda:
@gsemet this is probably something we can do. Can you explain your use case a little more?

gsemet:
My users need to belong at least to the video group to be able to start tensorflow-gpu. I think this is pretty legit use case, but this would involve setting a configuration such as:
```yaml
user_groups:
   - video
   - cuda
```
and upon user creation, it would `useradd -g video jupyter-myusername`

mylyu:
I just read the doc (https://tljh.jupyter.org/en/latest/topic/tljh-config.html#extra-user-groups), so currently we still need to add users to groups one by one, instead of automatically assigning them to groups on creation?

PR #382, opened 2019-06-20
Allow adding users to specific groups

GeorgianaElena:
 - [x] Add tests

@yuvipanda, I added support to add a user to a specific group, but I'm not sure if *UserCreatingSpawner* is the perfect place to do it.
What do you think?

closes https://github.com/jupyterhub/the-littlest-jupyterhub/issues/349

### DIFF
--- a/tljh/configurer.py
+++ b/tljh/configurer.py
@@ -27,7 +27,7 @@ default = {
         'allowed': [],
         'banned': [],
         'admin': [],
-        'groups': {}
+        'extra_user_groups': {}
     },
     'limits': {
         'memory': None,
@@ -175,7 +175,7 @@ def update_usergroups(c, config):
     Set user groups
     """
     users = config['users']
-    c.UserCreatingSpawner.user_groups = users['groups']
+    c.UserCreatingSpawner.user_groups = users['extra_user_groups']
 
 
 def update_limits(c, config):
--- a/tljh/configurer.py
+++ b/tljh/configurer.py
@@ -27,6 +27,7 @@ default = {
         'allowed': [],
         'banned': [],
         'admin': [],
+        'groups': {}
     },
     'limits': {
         'memory': None,
@@ -93,6 +94,7 @@ def apply_config(config_overrides, c):
 
     update_auth(c, tljh_config)
     update_userlists(c, tljh_config)
+    update_usergroups(c, tljh_config)
     update_limits(c, tljh_config)
     update_user_environment(c, tljh_config)
     update_user_account_config(c, tljh_config)
@@ -166,6 +168,14 @@ def update_userlists(c, config):
     c.Authenticator.whitelist = set(users['allowed'])
     c.Authenticator.blacklist = set(users['banned'])
     c.Authenticator.admin_users = set(users['admin'])
+
+
+def update_usergroups(c, config):
+    """
+    Set user groups
+    """
+    users = config['users']
+    c.UserCreatingSpawner.user_groups = users['groups']
 
 
 def update_limits(c, config):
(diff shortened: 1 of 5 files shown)


tljh/configurer.py
```diff
@@ -27,6 +27,7 @@
         'allowed': [],
         'banned': [],
         'admin': [],
+        'groups': {}
```
yuvipanda: Since this is only for users, can we call it something like extra_user_groups or something? 

Other than that this looks great :)

tljh/jupyterhub_config.py
```diff
@@ -13,12 +13,16 @@
...
 
+from traitlets import Any
+
 class UserCreatingSpawner(SystemdSpawner):
     """
     SystemdSpawner with user creation on spawn.
 
     FIXME: Remove this somehow?
     """
+    user_groups = Any(config=True)
```
yuvipanda: This should be a `List` traitlet not an `Any`

tljh/configurer.py
GeorgianaElena: Sure, *extra_user_groups* makes more sense :)

tljh/jupyterhub_config.py
GeorgianaElena: I don't think it can be a List :(
Because *extra_user_groups* will look something like this:
```
{
        'users': {
            'extra_user_groups': {
                "group1": ["u1", "u2"],
                "group2": ["u3", "u4"]
            },
        ....
        }
}
```
and we define ``c.UserCreatingSpawner.user_groups = users['extra_user_groups']``,
*user_groups* will be a dict like:
```
{
    "group1": ["user1", "user2"],
    "group2": ["user3", "user4"]
}
```
Unfortunately, I'm not sure there is a way to add to *UserCreatingSpawner.user_groups* only the list of groups that *user.name* must be added to :disappointed:.

I added some new commits where I defined user_groups to be a Dict traitlet instead of the generic Any.
Do you believe it's possible to have *UserCreatingSpawner* hold only the list of groups for that user and not the entire config, or is Dict good enough?

tljh/jupyterhub_config.py
```diff
@@ -13,12 +13,16 @@
...
 
+from traitlets import Dict
+
 class UserCreatingSpawner(SystemdSpawner):
     """
     SystemdSpawner with user creation on spawn.
 
     FIXME: Remove this somehow?
     """
+    user_groups = Dict(config=True)
```
minrk: Try this:

```suggestion
    user_groups = Dict(key_trait=Unicode(), value_trait=List(Unicode()), config=True)
```

This means that the keys will be validated as text (group names) and the values as lists of text (lists of group names). So if someone inputs by accident e.g.:

```python
c.UserCreatingSpawner.user_groups = {'group': 'user'}
```

the error should be caught at config time. The code as it is, would treat this the same as `{'group': ['u', 's', 'e', 'r']}` since a string is an iterable just like a list.

GeorgianaElena: Thanks for the suggestion @minrk, this is really nice. After 3 commits I finally managed to get it right :grimacing:

tljh/jupyterhub_config.py
yuvipanda: Ah, whoops. I missed this completely, apologies!

I think we shouldn't (right now at least) allow whitelisting like this, since we might not know all the user names at time of config (for most use cases, where the authenticator isn't FirstUseAuthenticator).

Instead, I think user_groups should be a list that's applied to *all* users. This satisfies the use case in #349, which is to help users get access to GPUs, etc. We can work on the per-user case when someone raises that for a specific use cases.

I hope that makes sense, and I'm so sorry it took me this long to respond!

GeorgianaElena: It makes perfectly sense. Thanks for taking the time to clarify this and don't worry about the delay :) Let me know if there's anything else I should change.

Issue #384, opened 2019-06-21
can't open file '/srv/src/tljh/cull_idle_servers.py': [Errno 2] No such file or directory


asimb123:
Hi,
I am seeing this error in log file after every 30 sec.
/opt/tljh/hub/bin/python3: can't open file '/srv/src/tljh/cull_idle_servers.py': [Errno 2] No such file or directory

I installed The littlest JupyterHub on AWS EC2.
Regards,
AB

yuvipanda:
Thanks for reporting this, @asimb123.

Do you have any insight into this, @GeorgianaElena

GeorgianaElena:
thanks for opening the issue, @asimb123.
@yuvipanda, I believe it's my fault, since I've hard-coded the path of the idle culler script (I have a development environment set up, not a regular install).
I've opened https://github.com/jupyterhub/the-littlest-jupyterhub/pull/386 where I'm running the idle culler as a module. Do you believe there is a better way of removing the hard-coded path? I don't know very well where everything should go (filesystem wise) and I might be missing something really obvious here)

yuvipanda:
Fixed by #386 

Thank you very much for getting to this, @GeorgianaElena!

PR #386, opened 2019-06-24
Run idle culler as a python module

GeorgianaElena:
Regarding https://github.com/jupyterhub/the-littlest-jupyterhub/issues/384, this PR removes the hard-coded path of the idle culler and runs the script as a python module.


yuvipanda:
Thanks, @GeorgianaElena!

PR #392, opened 2019-07-08
Import containers from collections.abc rather than collections

GeorgianaElena:


yuvipanda:
Thanks, @GeorgianaElena

Issue #263, opened 2019-01-29
Allow setting admin user password during install command


yuvipanda:
Right now, an attacker could race the admin to set their own password after creation but before the admin has set their password via FirstUseAuthenticator.

Allowing users to set admin password during set up should help with this particular threat vector.

Thanks to @jedbrown for raising this in https://github.com/jupyterhub/the-littlest-jupyterhub/issues/258

PR #395, opened 2019-07-10
Set admin password during install

GeorgianaElena:
Added the option to set the admin user password during install (closes https://github.com/jupyterhub/the-littlest-jupyterhub/issues/263).

If the installer is called with *--password* option, then a password will be requested for each admin user specified with *--admin**. If *--password* is not present, the script will behave like before.

@yuvipanda, can you please give me some advice about how/if should I test this? :eyes:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -73,6 +73,12 @@ jobs:
                     .circleci/integration-test.py run-test basic-tests test_hub.py test_install.py test_extensions.py
 
             - run:
+                name: Run admin tests
+                command: |
+                    .circleci/integration-test.py run-test --installer-args "--admin admin:admin" basic-tests test_admin_installer.py
+
+
+            - run:
                 name: Run plugin tests
                 command: |
                     .circleci/integration-test.py run-test \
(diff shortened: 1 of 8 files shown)


GeorgianaElena:
@yuvipanda, according to [the docs](http://tljh.jupyter.org/en/latest/topic/customizing-installer.html#adding-admin-users), in order to add multiple admins during install there should be something like: ```--admin a1 --admin a2```.
This didn't work for me (it only stored the last one). Doing instead ```--admin a1 a2``` works.
From argparse's docs it looks like there should be an *action="append"* in order to be able to use *--admin* multiple times.

yuvipanda:
w00t, thanks for working on this, @GeorgianaElena!

1. I think we should ask users for passwords in the form '--admin username:password --admin username:password', so we can match usernames to passwords easily. This re-uses how HTTP does basic auth usernames / passwords for syntax...
2. whoops re: append. Would love a PR for that.

What do you think?

GeorgianaElena:
Thanks for the feedback @yuvipanda :D
I will open a PR to add the *append*.
I chose not to have ```--admin:password``` because I feared that the password will be stored along with the install command in bash history and I think it is also visible when doing ```ps``` while the script is  running.
But I totally agree it is not ideal to have the installer request user input and if the concern above is not applicable here, I'm more then happy to have it as you suggested.

GeorgianaElena:
@yuvipanda, I've addressed your feedback, added tests and updated the docs. Let me know what you think :)

PR #399, opened 2019-07-11
Allow adding multiple admins during install 

GeorgianaElena:
Added the *append* action to the *--admin* option.


yuvipanda:
Awesome! Can you add a test for it too?

GeorgianaElena:
Just added one :) Thanks, @yuvipanda

Issue #402, opened 2019-07-15
'tljh-config set http.port' option does not work


ts-han:
Hi. I tried to set up The Littlest JupyterHub on a machine whose 80 port is already in use by apache2. I added the config to change the port of TLJH from 80 to 8080, and then reloaded/restarted the service. it seems that, however, the port change option does not work. How should I do in this case?

*admin@server:~$ sudo tljh-config show*
```
users:
  admin:
  - adminuser
http:
  port: 8080
user_environment:
  default_app: jupyterlab
```

*admin@server:~$ sudo tljh-config reload*
```
active
Hub reload with new configuration complete
```

*admin@server:~$ sudo service jupyterhub restart*

*admin@server:~$ sudo netstat -tnlp* (traefik still listen on 80 port)
```
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name
tcp        0      0 127.0.0.1:8099          0.0.0.0:*               LISTEN      4342/traefik
tcp        0      0 0.0.0.0:8787            0.0.0.0:*               LISTEN      931/rserver
tcp        0      0 127.0.0.1:53            0.0.0.0:*               LISTEN      889/acme-dns
tcp        0      0 127.0.0.53:53           0.0.0.0:*               LISTEN      707/systemd-resolve
tcp        0      0 0.0.0.0:22              0.0.0.0:*               LISTEN      1268/sshd
tcp        0      0 127.0.0.1:15001         0.0.0.0:*               LISTEN      21952/python3
tcp6       0      0 :::80                   :::*                    LISTEN      4342/traefik
tcp6       0      0 :::8081                 :::*                    LISTEN      889/acme-dns
tcp6       0      0 :::22                   :::*                    LISTEN      1268/sshd
```

Thank you in advance :)

GeorgianaElena:
Hi @ts-han! Thanks for reporting this.

*http.port* is part of the proxy's configuration, so in order for the change to take place, the proxy's configuration should be also reloaded with:
```
sudo tljh-config reload proxy
```

```reload_component``` can be used to either reload the hub or the proxy. If no arguments are provided, it defaults to hub.

ts-han:
Thank you, @GeorgianaElena!

`sudo tljh-config reload proxy` works well!

GeorgianaElena:
Happy it worked @ts-han! I'm going to close this issue now, but feel free to open a new one if you have any other question.

PR #403, opened 2019-07-15
Docs update: reload proxy after modifying the ports

GeorgianaElena:
Related to https://github.com/jupyterhub/the-littlest-jupyterhub/issues/402, it think we should make it clear that if the http/https ports are changed, the proxy should be reloaded.
 

yuvipanda:
Thanks, @GeorgianaElena

Issue #250, opened 2018-12-28
Upgrading to Python 3.7


fm75:
#247 Suggested that we could easily upgrade to 3.7.

I would suggest that we don't do that until [this TensorFlow issue](https://github.com/tensorflow/tensorflow/issues/20517) is resolved. It is relatively easily to install Tensorflow compatible with the current 3.6 interpreter.

Given its popularity today, I think it would be better to wait on this.

yuvipanda:
I think enough time has passed for 3.7 to be the default. However, I am not sure if we should automatically *upgrade* currently existing installations from 3.6 to 3.7. That must surely break things.

fm75:
I agree that making it the default should be good. According to the [ release notes](https://github.com/tensorflow/tensorflow/blob/master/RELEASE.md) TensorFlow now supports version 3.7 "on all operating systems".

shireenrao:
I tried to upgrade my python under `/opt/tljh/user/bin/python` using the command 
```
conda install python=3.7.2
```
I can confirm that python got installed fine, but looks like jupyterhub is using python under `/opt/tljh/hub/bin`. which is still at `3.6.7`. So my install is broken. How would one upgrade python under `/opt/tljh/hub/bin`? Here is a snippet from my logs

```
May 24 02:59:52 systemd[1]: Started jupyterhub.service.
May 24 02:59:53 python3[931]: [I 2019-05-24 02:59:53.417 JupyterHub app:1673] Using Authenticator: firstuseauthenticator.firstuseauthenticator.FirstUseAuthenti
May 24 02:59:53 python3[931]: [I 2019-05-24 02:59:53.417 JupyterHub app:1673] Using Spawner: builtins.UserCreatingSpawner
May 24 02:59:53 python3[931]: [I 2019-05-24 02:59:53.418 JupyterHub app:1016] Loading cookie_secret from /opt/tljh/state/jupyterhub_cookie_secret
May 24 02:59:53 python3[931]: [I 2019-05-24 02:59:53.478 JupyterHub app:1201] Not using whitelist. Any authenticated user will be allowed.
May 24 02:59:53 python3[931]: [W 2019-05-24 02:59:53.503 JupyterHub app:1462] abc appears to have stopped while the Hub was down
May 24 02:59:53 python3[931]: [W 2019-05-24 02:59:53.505 JupyterHub app:1462] xyz appears to have stopped while the Hub was down
[...]

PR #433, opened 2019-09-18
Make Python3.7 the default

GeorgianaElena:
This PR follows the steps in https://github.com/jupyterhub/the-littlest-jupyterhub/issues/250#issuecomment-511085121 to use Python3.7 with fresh TLJH installs, but doesn't upgrade the Python version for existing installs.

### DIFF
--- a/docs/howto/env/user-environment.rst
+++ b/docs/howto/env/user-environment.rst
@@ -153,3 +153,57 @@ The most common & portable way to fix this when using ``ssh`` is:
 .. code-block:: bash
 
    sudo PATH=${PATH} conda install -c conda-forge gdal
+
+
+Upgrade to a newer Python version
+=================================
+
+All new TLJH installs use miniconda 4.7.10, which comes with a Python 3.7
+environment for the users. The previously TLJH installs came with miniconda 4.5.4,
+which meant a Python 3.6 environment.
+
+To upgrade the Python version of the user environment, one can:
+
+*  **Start fresh on a machine that doesn't have TLJH already installed.**
+
+   See the :ref:`installation guide <install/installing>` section about how to install TLJH.
+
+*  **Upgrade Python manually.**
+
+   Because upgrading Python for existing installs can break packages alaredy installed
+   under the old Python, upgrading your current TLJH installation, will NOT upgrade
+   the Python version of the user environment, but you may do so manually.
+
+   **Steps:**
+
+   1. Activate the user environment, if using ssh.
+      If the terminal was started with JupyterHub, this step can be skipped:
+
+      .. code-block:: console
+
+         source /opt/tljh/user/bin/activate
+
+   2. Get the list of currently installed pip packages (so you can later install them under the
+      new Python):
+
+      .. code-block:: console
+
+         pip freeze > pip_pkgs.txt
+
+   3. Update all conda installed packages in the environment:
+
+      .. code-block:: console
+
+         sudo conda update --all
+
+   4. Update Python version:
+
+      .. code-block:: console
+
+         conda install python=3.7
+
+   5. Install the pip packages previously saved:
+
+      .. code-block:: console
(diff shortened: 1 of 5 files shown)


betatim:
Reading this PR I realised there are two Python versions related to TLJH. The Python version in the conda environment used by users and the Python version installed on the host.

The version on the host is what is used to install and run the hub itself. The version in the conda env is what users see (their kernels run in this env).

Then there is also the version of conda that is used.

Is that also your summary? (I am new to the internals of TLJH.)

---

I thought that this PR was about updating the version that users see for their kernels. Looks like you are doing all three at the same time. Being a simple minded person I would try and upgrade one of them at a time (starting with the one that is seen by users). That way the number of things that can break and reasons for it breaking are kept to a minimum. What do you think?

GeorgianaElena:
@betatim, you summarized the setup perfectly.

---

> I would try and upgrade one of them at a time (starting with the one that is seen by users). That way the number of things that can break and reasons for it breaking are kept to a minimum. What do you think?

 This PR doesn't bump Python and conda for existing installs, so it shouldn't break anything. I tested this on my TLJH DO deployment and the Python and conda versions remained unchanged.

This PR just makes Python3.7 the default Python version used in TLJH (in the hub environment) and upgrades conda to a newer version, which also upgrades Python to 3.7 version in the user env (ref: https://github.com/jupyterhub/the-littlest-jupyterhub/pull/336).

Not sure if upgrading one at a time would be easier or less error prone (In the issue referenced in my initial comment there were people that tried to upgrade python separately and ended up with a broken env).

P.S> The tests were failing because I have forgotten to bump the conda version in tests also.

yuvipanda:
Thanks for working on this, and for being so so patient, @GeorgianaElena!

There are two different Pythons:

1. Python in the venv that runs the hub, running system python (currently 3.6)
2. Default Python that comes with miniconda

I *think* we shouldn't touch the venv right now - there isn't any reason to, and we can bump it when we move from Ubuntu Bionic to the next LTS release when it comes out.

For the user environment, I think we should install newer miniconda (and with it, newer python), and not actually try to upgrade python of existing installs. This can break packages people have installed, esp with pip. If folks wanna upgrade their python, they can do so manually with a conda command I think.

So in summary, I think we should:

1. Not update the venv python version
2. Install newer miniconda with newer python only on newer installs
3. Leave older miniconda installations alone, but provide documentation on how to upgrade python there

How does this sound?

GeorgianaElena:
Thanks @yuvipanda, @betatim for the review :heart: 

I added new commits that:
* leave the venv Python version as it is
* check if miniconda 4.5.4 or 4.7.10 exist (and installs 4.7.10 if no miniconda)
* add a new docs section that document how to manually upgrade Python for existing TLJH installations
   However, I'm not sure if the steps I provided can lead to trouble (I tried them and managed to upgrade) or if all of them are needed.

docs/howto/env/user-environment.rst
```diff
@@ -153,3 +153,57 @@ The most common & portable way to fix this when using ``ssh`` is:
...
+
+      .. code-block:: bash
+
+         sudo PATH=${PATH} conda install python=3.7
+
+   5. Install the pip packages previously saved:
+
+      .. code-block:: bash
+
+         pip install pip_pkgs.txt
```
yuvipanda:
```suggestion
         pip install -r pip_pkgs.txt
```

yuvipanda:
@GeorgianaElena awesome! Minor suggestions, happy to merge after

betatim:
Thanks for the work!

To clarify the "update one at a time" comment: I was thinking of two PRs, so that we can deal with build failures from changing things one at a time instead of having the combined build failures of two changes.

yuvipanda:
Awesome! Thanks for your patience, @GeorgianaElena!

Issue #431, opened 2019-09-17
Systemd issue


GeorgianaElena:
As pointed out in https://github.com/jupyterhub/the-littlest-jupyterhub/issues/408 and https://github.com/jupyterhub/the-littlest-jupyterhub/issues/237, having ``ProtectHome=tmpfs`` raises some issues:
-  systemd version in Ubuntu Bionic is 237 and ``ProtectHome=tmpfs`` isn't recognized as a valid option, which causes *ProtectHome* to default to *no*.
  (check this with ``systemctl show jupyterhub.service -p ProtectHome`` shows: *ProtectHome=no*)
  Changing ``ProtectHome=yes``(jupyterhub process should have no access to home directories) results in failing spawns because the users home directories don't get created.
- systemd version in [Ubuntu 18.10 is 239](https://packages.ubuntu.com/cosmic/systemd), which supports ``ProtectHome=tmpfs``, but as pointed out in https://github.com/jupyterhub/the-littlest-jupyterhub/issues/237, this doesn't work as expected and users home directories don't get created.

Reading through [systemd's man](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#ProtectHome=), it is my understanding that:
1. having ``ProtectHome=yes``(Ubuntu 18.04) in *jupyterhub.service* makes */home* inaccessible and empty for processes invoked by the jupyterhub unit. This means that when the [spawner is trying to add the user and create its home directory](https://github.com/jupyterhub/the-littlest-jupyterhub/blob/master/tljh/jupyterhub_config.py#L35), because it's being invoked by the jupyterhub unit (thus, */home* is not accessible), it fails with:  
```
Previous spawn for abcd failed: Command '['useradd', '--create-home', 'jupyter-abcd']' returned non-zero exit status 12.
```
2. having ``ProtectHome=tmpfs`` (Ubuntu 18.10)  in *jupyterhub.service*  a temporary file system is mounted on */home* in read-only mode. Being read-only, when the spawner is trying to create the user's home directory */home/jupyter-user* will result in a failure (though I don't understand why an error isn't shown).

I *think* that:
- on Ubuntu 18.04 everything works because ``ProtectHome=tmpfs`` is not recognized and it becomes ``ProtectHome=no`` giving to the jupyterhub unit access to the home directory and thus it's able to create sub-directories in */home* for the users)
[...]

Issue #408, opened 2019-07-30
Systemd config does not validate under Ubuntu Bionic


stormerider:
```
Jul 30 20:56:50 dev-ml systemd[1]: /etc/systemd/system/jupyterhub.service:12: Failed to parse protect home value, ignoring: tmpfs
```
This is from the jupyterhub service file:
```
root@dev-ml:/opt/tljh/config# cat /etc/systemd/system/jupyterhub.service 
# Template file for JupyterHub systemd service
# Uses simple string.format() for 'templating'
[Unit]
# Traefik must have successfully started *before* we launch JupyterHub
Requires=traefik.service
After=traefik.service

[Service]
User=root
Restart=always
# jupyterhub process should have no access to home directories
ProtectHome=tmpfs
WorkingDirectory=/opt/tljh/state
# Protect bits that are normally shared across the system
PrivateTmp=yes
PrivateDevices=yes
ProtectKernelTunables=yes
ProtectKernelModules=yes
Environment=TLJH_INSTALL_PREFIX=/opt/tljh
# Run upgrade-db before starting, in case Hub version has changed
# This is a no-op when no db exists or no upgrades are needed
ExecStart=/opt/tljh/hub/bin/python3 -m jupyterhub.app -f /opt/tljh/hub/lib/python3.6/site-packages/tljh/jupyterhub_config.py --upgrade-db

[Install]
# Start service when system boots
WantedBy=multi-user.target
root@dev-ml:/opt/tljh/config# 
```
From my research, ProtectHome=tmpfs was introduced in [version 238](https://www.phoronix.com/scan.php?page=news_item&px=Systemd-238-Released) of systemd. Ubuntu Bionic is still on version 237.

feria35:
Hello @stormerider or anyone who has that issue.
My VPS is on Ubuntu 18.04 Server (by OVH) Bionic.
I had exactly the problem you described and I noticed it only happened when NginX was already installed, I mean before TLJH install. Stopping the service NginX is not enough to get rid of that error at TLJH's install.
So, from nowadays, I install **TLJH before NginX**, Let's Encrypt, PHP etc ..  and it's OK now.
I can't give you any explaination, but it might help you.

stormerider:
That's interesting. I do have nginx installed. and it probably was before TLJH. I still think it is a bug; it's trying to use a feature that Bionic doesn't support.

Issue #237, opened 2018-11-22
TLJH does not automatically create home directories for new users on Ubuntu Server 18.10


owah:
Once tljh was up and running ([after I changed the installer locally to use a newer node version](https://github.com/jupyterhub/the-littlest-jupyterhub/issues/236)), I couldn't start any notebooks, because the spawner process would time out. I have looked into the logging and came upon the following:

```
sudo journalctl -u jupyter-testuser:

Nov 22 12:09:20 blockchainlab systemd[1]: Started /bin/bash -c cd /home/jupyter-testuser && exec jupyterhub-singleuser --port=36531.
Nov 22 12:09:20 blockchainlab systemd[20897]: jupyter-testuser.service: Changing to the requested working directory failed: No such file or directory
Nov 22 12:09:20 blockchainlab systemd[20897]: jupyter-testuser.service: Failed at step CHDIR spawning /bin/bash: No such file or directory
Nov 22 12:09:20 blockchainlab systemd[1]: jupyter-testuser.service: Main process exited, code=exited, status=200/CHDIR
Nov 22 12:09:20 blockchainlab systemd[1]: jupyter-testuser.service: Failed with result 'exit-code'.
```

I have manually created a home directory for the user, assigned the proper permissions and then restarted the service. After doing so I was able to launch a Python 3 notebook.

yuvipanda:
Thanks for opening this, @owah! Can you provide logs from JupyterHub itself (`sudo journalctl -u jupyterhub`)? That should hopefully provide output on why the initial adduser command is failing.

owah:
hi @yuvipanda, this is the weird thing, the command does not fail! I've checked the logs and all I see is:
```
Nov 23 21:50:11 juphubtest useradd[3081]: new group: name=jupyter-testuser, GID=1004
Nov 23 21:50:11 juphubtest useradd[3081]: new user: name=jupyter-testuser, UID=1002, GID=1004, home=/home/jupyter-testuser, shell=/bin/sh
```

yuvipanda:
@owah interesting. is this repeatable? I wonder if adduser behavior changed in 18.l0.

owah:
@yuvipanda it is.

What I have tried so far:
replace `subprocess.check_call()` with `os.system("useradd --create-home " + username)` and `os.system("useradd --create-home hardcoded-name")`

In both cases, the user gets added to `/etc/passwd` but no home-dir. is created.

[...]

PR #435, opened 2019-09-19
Disable ProtectHome=tmpfs

GeorgianaElena:
This disables ProtectHome for jupyterhub process and enables it for the traefik process (ref: https://github.com/jupyterhub/the-littlest-jupyterhub/issues/431).

Should fix https://github.com/jupyterhub/the-littlest-jupyterhub/issues/431, https://github.com/jupyterhub/the-littlest-jupyterhub/issues/408 and https://github.com/jupyterhub/the-littlest-jupyterhub/issues/237.


minrk:
Thanks!

Issue #445, opened 2019-10-10
`apt.add_source` keeps adding lines


willirath:
On one of my Ubuntu deployments, I've found a ` /etc/apt/sources.list.d/nodesource.list` containing the following:

```
deb https://deb.nodesource.com/node_10.x bionic maindeb https://deb.nodesource.com/node_10.x bionic maindeb https://deb.nodesource.com/node_10.x bionic maindeb https://deb.nodesource.com/node_10.x bionic maindeb https://deb.nodesource.com/node_10.x bionic maindeb https://deb.nodesource.com/node_10.x bionic maindeb https://deb.nodesource.com/node_10.x bionic maindeb https://deb.nodesource.com/node_10.x bionic maindeb https://deb.nodesource.com/node_10.x bionic main
```

If I remember rightly, we had connectivity problems when setting up the machine, and re-ran the TLJH setup multiple times. My understanding of https://github.com/jupyterhub/the-littlest-jupyterhub/blob/master/tljh/apt.py#L21_L36 is that it should prevent adding the source if it's already contained. But it does seem to fail at that.

PR #449, opened 2019-10-21
Write out deb line only if it already doesn't exist

GeorgianaElena:
Should fix https://github.com/jupyterhub/the-littlest-jupyterhub/issues/445


yuvipanda:
Awesome, thanks @georgianaelena

PR #453, opened 2019-10-24
Add hook for new users

### LINKED ISSUE #451: Add new-user-hook
Hi,

I would like a hook, which runs code after a new user was created. In my case I would use it to configure a specific database access with the user's credentials after the first login. 

What are your thoughts on this? Any counter arguments?

edit: just saw that the plugin-manager is specific to installer.py. Do you think it would be easy to adapt for the new-user-creation method?

jkfm:
 - [ ] Add tests

So I plugged the hook into ensure_user, which works locally. I did not yet write a test, because I am unsure about where would be the best location to test this functionality (and I did not want to break any existing test). Maybe you could point this out to me. 

I think adding documentation would not be needed, as the hooks are referenced as a python file in the docs.

 

(Closes #451 )

### DIFF
--- a/tljh/user.py
+++ b/tljh/user.py
@@ -8,14 +8,8 @@ import pwd
 import subprocess
 from os.path import expanduser
 
-import pluggy
-
-from tljh import hooks
-
 # Set up plugin infrastructure
-pm = pluggy.PluginManager('tljh')
-pm.add_hookspecs(hooks)
-pm.load_setuptools_entrypoints('tljh')
+from tljh.utils import get_plugin_manager
 
 
 def ensure_user(username):
@@ -30,6 +24,8 @@ def ensure_user(username):
     except KeyError:
         # User doesn't exist, time to create!
         pass
+
+    pm = get_plugin_manager()
 
     subprocess.check_call([
         'useradd',
--- a/tljh/user.py
+++ b/tljh/user.py
@@ -37,13 +37,13 @@ def ensure_user(username):
         username
     ])
 
-    pm.hook.tljh_new_user_create(username=username)
-
     subprocess.check_call([
         'chmod',
         'o-rwx',
         expanduser('~{username}'.format(username=username))
     ])
+
+    pm.hook.tljh_new_user_create(username=username)
 
 
 def remove_user(username):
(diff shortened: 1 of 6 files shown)


tljh/user.py
```diff
@@ -3,11 +3,20 @@
...
 import grp
+import pwd
 import subprocess
 from os.path import expanduser
 
+import pluggy
+
+from tljh import hooks
+
+# Set up plugin infrastructure
```
yuvipanda: We already instantiate a PluginManager object here: https://github.com/jupyterhub/the-littlest-jupyterhub/blob/master/tljh/jupyterhub_config.py#L72. users.py gets imported from the same place. I'd like us to have only one PluginManager object per process. Can you put it in utils.py maybe?

tljh/user.py
```diff
@@ -28,9 +37,11 @@ def ensure_user(username):
         username
     ])
 
+    pm.hook.tljh_new_user_create(username=username)
```
yuvipanda: We should probably move this to the end, after the chmod - in case they wanna undo the effects of the chmod.

tljh/user.py
jkfm: Good call!

tljh/user.py
jkfm: I agree as well

jkfm:
So I agree that there should be some tests, I'll try to find some time for them!

yuvipanda:
@jkfm it turns out I was wrong, and there *are* some plugin tests! `integration-tests/test_simplest_plugin.py`. The plugin it is testing is under `integration-tests/plugins/simplest`. We could maybe create a similar plugin for this and test it?

Apologies for the very delayed review...

jkfm:
No worries, am on vacation myself and will take a look in a couple of weeks.

So you are saying, that an integration test would suffice here?

Regards,
Johannes

Yuvi Panda <notifications@github.com> schrieb am Mi., 13. Nov. 2019, 00:25:

yuvipanda:
@jkfm yep! An integration test would be good enough

jkfm:
Hey @yuvipanda , do you know if the pipeline fails due to my added test? Because the specific error message points to the conda-package.

jkfm:
Friendly ping @yuvipanda :)

yuvipanda:
I think the failure is unrelated, @jkfm. I've merged this now - apologies for the delay!

jkfm:
Thanks!

PR #457, opened 2019-10-28
switch base environment to requirements file

minrk:
requirements file is easier to work with and update than inlined strings

In the future, we should be doing `pip freeze` and using the frozen result, like we do (with conda) in repo2docker because pinning only direct dependencies is a way to guarantee broken environments in the future (see the pinning-down of tornado as one example).

Several dependencies are updated in the process (jupyterlab 1.1, notebook 6.0, etc.)

### DIFF
--- a/integration-tests/test_extensions.py
+++ b/integration-tests/test_extensions.py
@@ -13,14 +13,14 @@ def test_serverextensions():
     ], stderr=subprocess.PIPE)
 
     extensions = [
-        'jupyterlab 1.2.6',
-        'nbgitpuller 0.6.1',
-        'nteract_on_jupyter 2.0.7',
+        'jupyterlab 1.2.',
+        'nbgitpuller 0.7.',
+        'nteract_on_jupyter 2.1.',
         'nbresuse '
     ]
 
     for e in extensions:
-        assert '{} \x1b[32mOK\x1b[0m'.format(e) in proc.stderr.decode()
+        assert e in proc.stderr.decode()
 
 def test_nbextensions():
     """
--- a/tljh/installer.py
+++ b/tljh/installer.py
@@ -272,22 +272,10 @@ def ensure_user_environment(user_requirements_txt_file):
         'conda==' + conda_version
     ])
 
-    conda.ensure_pip_packages(USER_ENV_PREFIX, [
-        # JupyterHub + notebook package are base requirements for user environment
-        'jupyterhub==1.0.0',
-        'notebook==5.7.8',
-        # Install additional notebook frontends!
-        'jupyterlab==1.2.6',
-        'nteract-on-jupyter==2.0.7',
-        # nbgitpuller for easily pulling in Git repositories
-        'nbgitpuller==0.6.1',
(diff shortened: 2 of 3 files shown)


yuvipanda:
@minrk totally agree re: freeze.

Failing with:

```
Ran /opt/tljh/user/bin/jupyter labextension install @jupyterlab/hub-extension @jupyter-widgets/jupyterlab-manager with exit code 1
-/|\-/|\-/|\-/|Building jupyterlab assets (build:prod:minimize)
-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/|\-/An error occured.
RuntimeError: JupyterLab failed to build
See the log file for details:  /tmp/jupyterlab-debug-vdawix52.log

```

With 1.1 we don't need hub-extension either.

GeorgianaElena:
I ran into this problem (jupyterlab failing to build) on my TLJH deployment on DO when I upgraded jupyterlab manually from 0.35.4 to 1.2.5.
From the logs, it looked like the failure was related to https://github.com/jupyterlab/jupyterlab/issues/7175. After increasing the memory through the env var ``NODE_OPTIONS``, the build went through.

I tried to solve the failing tests in this PR it in the same manner but with no luck :( Either there's something missing or the tests are failing from a whole different reason.

yuvipanda:
@GeorgianaElena I wonder if this needs a rebase - we fixed some other errors in master after this PR I think.

letianw91:
> jupyterlab/jupyterlab#7175

I think `@jupyterlab/hub-extension` is obsolete in Jupyterlab 1.0.0 and above.
See [release-notes](https://jupyterlab.readthedocs.io/en/stable/getting_started/changelog.html) (search `@jupyterlab/hub-extension`) which might have caused the build error.

letianw91:
Another reason for flaky build is that Webpack minify process is resource intensive, which might cause flaky build on pipelines with smaller machines. add `--minimize=False` to the labextension install seems to solve the issue for me. Will raise a PR later

GeorgianaElena:
I've rebased the PR and now it contains @letianw91 's fix :rocket:

yuvipanda:
Awesome, ty @GeorgianaElena!

PR #468, opened 2019-11-21
Switch to pandas theme

yuvipanda:
Ref https://github.com/jupyterhub/team-compass/issues/225. I like
the new theme as well.

The top level headers still need to be sorted out.


yuvipanda:
image

is how it looks now, but it shouldn't look like that.

choldgraf:
In my case, I generally needed to:

1. Define which pages I wanted as top-level headings
2. For each one, make sure there was an "index" page for that heading. In many cases this corresponded to folders in the `source` folder, so each folder had an `index.rst` file. But in other repositories with a flat structure, I just made an `index-heading1.rst` and `index-heading2.rst` file etc.
3. In each index page, its `.. toctree::` should point to the pages for that sub-heading.
4. On the *main* index page, it should *only* have a `.. toctree::` that points to the top-level heading index pages.

Does that help?

choldgraf:
hey @yuvipanda - I hope that it is OK with you, but I just pushed a commit to your branch that fixes up some of the page naming and index issues, and also adds a little logo for the site. Lemme know what you think - feel free to change anything, or to ask for any changes.

choldgraf:
(btw - it was super easy to reorganize the index files because these docs are **awesomely** organized)

PR #335, opened 2019-05-19
Limit memory available in integration tests

yuvipanda:
- Run all integration tests with a 768MB memory limit,
  so we know when we're using too much RAM
- Document this memory requirement in our docs.

Fixes #257

 - [x] Add / update documentation
 - [x] Add tests

PR #479, opened 2020-01-13
Fix failing integration tests

GeorgianaElena:
Apparently the 768MB memory limit is not enough anymore and it caused the integration tests failure.
Increasing it to 1GB seems to solve the issues (at least for now).

ref: https://github.com/jupyterhub/the-littlest-jupyterhub/issues/257 and https://github.com/jupyterhub/the-littlest-jupyterhub/pull/335

P.S: Remaining things to do for the tests to pass: find out why the test in https://github.com/jupyterhub/the-littlest-jupyterhub/pull/453 fails.


yuvipanda:
Awesome!!! Thank you for digging in and taking care of this :)

PR #491, opened 2020-01-25
[MRG] Fix integration test for new pip

### LINKED ISSUE #490: Failing integration tests
I think the first new feature in https://pip.pypa.io/en/stable/news/#features is the reason that the integration tests now fail.

The way I read the test that fails is that it checks that you can't install packages as a "user". However the new version of `pip` will switch to "user install" mode when it detects that the global directory isn't writeable. 

i think the thing for us to test now is that `pip install` as a user installs the package to a location that isn't "the system location" (for pip 20.x and up) or fails (old versions of pip).

Pretty sure that in either case nothing in TLJH has broken, it is "just" our test that users can't install things globally that is getting a false positive.

betatim:
This is an attempt to fix the integration test. Closes #490.

The TL;DR is that the [newest version of pip](https://pip.pypa.io/en/stable/news/#features) will switch to the behaviour of `pip install --user ...` when you run `pip install ...` without being able to write to the global package directory. This means our integration test started to fail because the install we were doing now succeeded. However we were testing if a user could install things to the global directory or not. This means the test failing was a false positive (the install succeeded but not because the user could write to the global dir).

This PR makes it so that the `pip` command used in the test won't enable `--user` if it can't write to the global directory. Instead it will fail (as it used to).

An alternative would be to check if the pip command fails and if it succeeds check the location the package got installed to.

### DIFF
--- a/integration-tests/test_install.py
+++ b/integration-tests/test_install.py
@@ -207,6 +207,7 @@ def test_pip_upgrade(group, allowed):
                 "-m",
                 "pip",
                 "install",
+                "--no-user",
                 "--ignore-installed",
                 "--no-deps",
                 "testpath==0.3.0",
@@ -218,6 +219,7 @@ def test_pip_upgrade(group, allowed):
             [python, "-m", "pip", "install", "--upgrade", "testpath"],
             preexec_fn=partial(setgroup, group),
         )
+
 
 def test_symlinks():
     """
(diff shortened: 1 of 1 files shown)


GeorgianaElena:
Yaay! Thanks for fixing this @betatim :sparkles:

Issue #488, opened 2020-01-23
How to disable use of TLS version 1.0


aye-aye-aye:
Our tljh has been flagged for a few vulnerability issues in a recent Qualys scan. We'd like to disable use of TLS v1.0 and v1.1 and some legacy block ciphers.

From reading the docs, I think we have to adjust the settings in /opt/tljh/state/traefik.toml, but I can't get it to work when following the advice laid out here: https://docs.traefik.io/https/tls/#tls-options

We are using manual HTTPS with an existing self-signed key and cert as described here: http://tljh.jupyter.org/en/latest/howto/admin/https.html#manual-https-with-existing-key-and-certificate

Has anyone any advice on how to disable TLS v1.0-1, or set minimum version as v1.2?

GeorgianaElena:
Hey @aye-aye-aye! Thanks for reporting this issue.
At the moment, the TLS version is not configurable, but your use-case makes a lot of sense and I believe we should make it configurable.

Making the TLS version configurable will require to:
* add [tls options](https://docs.traefik.io/https/tls/#tls-options) to the default config [here](https://github.com/jupyterhub/the-littlest-jupyterhub/blob/7d30501fc59567c33890c215354869d16df698d0/tljh/configurer.py#L19) (maybe even make TLS v1.2 be the default version)
* read the tls options from the default config and add them to the [traefik.toml template](https://github.com/jupyterhub/the-littlest-jupyterhub/blob/909dd70af2c8ad8b20344a6b2ccf752cee3811ce/tljh/traefik.toml.tpl)
* add a test [here](https://github.com/jupyterhub/the-littlest-jupyterhub/blob/master/tests/test_configurer.py)
* document how the TLS version can be changed using `tljh-config set`

Would be lovely if you could make a PR :yellow_heart: I'll be here to answer any questions.

yuvipanda:
I spent some time getting the z2jh Traefik HTTPS setup to be as good as possible, balancing secure HTTPS vs support for older clients. We should be able to just steal that code - https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/jupyterhub/templates/proxy/autohttps/configmap.yaml#L65. This requires we use traefik 2.0 however...

yuvipanda:
I don't think we need to make TLS version configurable - we should just set up for what we count as good defaults, which in this case would use TLS 1.2 and disable some older versions...

aye-aye-aye:
[...]

PR #498, opened 2020-01-31
Set tls 1.2 to be the min version

GeorgianaElena:
This sets TLS 1.2 to be the minimum tls version for the https entry point.

traefik 1.7 docs ref: https://docs.traefik.io/v1.7/configuration/entrypoints/#specify-minimum-tls-version

closes https://github.com/jupyterhub/the-littlest-jupyterhub/issues/488

### DIFF
--- a/tests/test_traefik.py
+++ b/tests/test_traefik.py
@@ -69,7 +69,7 @@ def test_letsencrypt_config(tljh_dir):
 
     assert cfg["entryPoints"] == {
         "http": {"address": ":80", "redirect": {"entryPoint": "https"}},
-        "https": {"address": ":443", "tls": {}},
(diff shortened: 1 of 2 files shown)


yuvipanda:
I looked at https://caniuse.com/#feat=tls1-2 to see what browsers we might lose from this. IE10 and below seem the only casualities, and I'm ok with that!

yuvipanda:
Thanks, @GeorgianaElena!

aye-aye-aye:
Could you please advise if it is possible to apply this enhancement to a running tljh instance without having to run the bootstrap? I am encountering the same install issues as detailed in https://github.com/jupyterhub/the-littlest-jupyterhub/issues/502, however we are under pressure from our internal VM team to remove support for TLS v1.1 from our instance asap. Any help would be greatly appreciated!

PR #499, opened 2020-02-01
Upgrade jupyterlab to 1.2.6

letianw91:
This PR upgrades Jupyterlab to the latest version of 1.2.6 as of Feb 1st 2020. Build and integration tests passes locally. 

- `jupyterlab-hubextension` removed because this has been integrated to Jupyterlab 1.0.0+ 
- Added `--minimize=False` option to Jupyterlab installation to avoid flaky build on low resource CI pipelines

 - [X] Add tests

### DIFF
--- a/tljh/installer.py
+++ b/tljh/installer.py
@@ -170,14 +170,13 @@ def ensure_jupyterlab_extensions():
     Install the JupyterLab extensions we want.
     """
     extensions = [
-        '@jupyterlab/hub-extension',
-        '@jupyter-widgets/jupyterlab-manager'
+        '@jupyter-widgets/jupyterlab-manager@1.1' # for jupyterlab 1.2.x
     ]
     utils.run_subprocess([
         os.path.join(USER_ENV_PREFIX, 'bin/jupyter'),
         'labextension',
         'install'
-    ] + extensions)
+    ] + extensions + ['--minimize=False']) # webpack minimization might cause flaky test build
 
 
 def ensure_jupyterhub_package(prefix):
@@ -263,7 +262,7 @@ def ensure_user_environment(user_requirements_txt_file):
         'jupyterhub==1.0.0',
         'notebook==5.7.8',
         # Install additional notebook frontends!
-        'jupyterlab==0.35.4',
+        'jupyterlab==1.2.6',
         'nteract-on-jupyter==2.0.7',
         # nbgitpuller for easily pulling in Git repositories
         'nbgitpuller==0.6.1',
--- a/tljh/installer.py
+++ b/tljh/installer.py
@@ -172,11 +172,26 @@ def ensure_jupyterlab_extensions():
     extensions = [
         '@jupyter-widgets/jupyterlab-manager@1.1' # for jupyterlab 1.2.x
     ]
+    install_options = [
+        '--no-build'   # do not build extension at install time. Will build later
+    ]
     utils.run_subprocess([
         os.path.join(USER_ENV_PREFIX, 'bin/jupyter'),
         'labextension',
         'install'
-    ] + extensions + ['--minimize=False']) # webpack minimization might cause flaky test build
+    ] + extensions + install_options)
+    
+    # Build all the lab extensions in one go using jupyter lab build command
+    build_options = [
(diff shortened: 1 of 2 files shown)


letianw91:
```srv/src/integration-tests/test_install.py:143: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

group = 'jupyterhub-admins', path = '/opt/tljh/state/jupyterhub.sqlite-journal'

    def permissions_test(group, path, *, readable=None, writable=None, dirs_only=False):
        """Run a permissions test on all files in a path path"""
        # start a subprocess and become nobody:group in the process
        pool = ProcessPoolExecutor(1)
        pool.submit(setgroup, group)
    
        def access(path, flag):
            """Run access test in subproccess as nobody:group"""
            return pool.submit(os.access, path, flag).result()
    
        total_tested = 0
    
        failures = []
    
        # walk the directory and check permissions
        for root, dirs, files in os.walk(path):
            to_test = dirs
            if not dirs_only:
                to_test += files
            total_tested += len(to_test)
            for name in to_test:
                path = os.path.join(root, name)
                if os.path.islink(path):
                    # skip links
                    continue
>               st = os.lstat(path)
E               FileNotFoundError: [Errno 2] No such file or directory: '/opt/tljh/state/jupyterhub.sqlite-journal'```

Getting this error in the CI build. I feel like this is a race-condition/flaky build problem: Some processes resulted in a temp sqlite file that existed at list folder time but have disappeared at permission check time.

File: tljh/installer.py [hunk 1]
```diff
@@ -170,14 +170,13 @@ def ensure_jupyterlab_extensions():
     Install the JupyterLab extensions we want.
     """
     extensions = [
-        '@jupyterlab/hub-extension',
```
betatim: Why are you removing this extension?

betatim:
It feels weird to add a special option to the installer just to get it to pass on the CI. However there is https://github.com/jupyterlab/jupyterlab/issues/7180 which also does this and (found via that issue) the https://github.com/jupyter/docker-stacks/ images also use this option.

With multiple lab extensions you can use something like the following to only build things once instead of for each extension you install:
```
jupyter labextension install jupyterlab-plotly@1.5.0 --no-build
jupyter labextension install plotlywidget@1.5.0 --no-build
# Added --minimize=False per https://github.com/jupyterlab/jupyterlab/issues/7180
jupyter lab build --minimize=False --dev-build=False
```

betatim:
I don't think we run our test suite in parallel so the flakeyness must come from somewhere else :-/

tljh/installer.py
letianw91: This plugin has been integrated to Juypterhub proper since 1.0.0: https://jupyterlab.readthedocs.io/en/stable/getting_started/changelog.html

>JupyterHub Integration
We now include the JupyterHub extension in core JupyterLab, so you no longer need to install @jupyterlab/hub-extension. (#6451, #6428)

letianw91:
> It feels weird to add a special option to the installer just to get it to pass on the CI. However there is [jupyterlab/jupyterlab#7180](https://github.com/jupyterlab/jupyterlab/issues/7180) which also does this and (found via that issue) the https://github.com/jupyter/docker-stacks/ images also use this option.
> 
> With multiple lab extensions you can use something like the following to only build things once instead of for each extension you install:
> 
> ```
> jupyter labextension install jupyterlab-plotly@1.5.0 --no-build
> jupyter labextension install plotlywidget@1.5.0 --no-build
> # Added --minimize=False per https://github.com/jupyterlab/jupyterlab/issues/7180
> jupyter lab build --minimize=False --dev-build=False
> ```

Yep, I agree it is weird to do this. The suggested method is good, and is actually how I tend to build lab extensions myself. Will update to use this method.

tljh/installer.py
betatim: Aha! I didn't know that. Every day is a school day :)

betatim:
Nice work, merging!

Issue #502, opened 2020-02-03
Install issues on Ubuntu 18.04


christasich:
I'm running into an error when trying to install TLJH on my personal server (Ubuntu Server 18.04.04 LTS) I start by issuing this command on my admin account.

`curl https://raw.githubusercontent.com/jupyterhub/the-littlest-jupyterhub/master/bootstrap/bootstrap.py | sudo -E python3 - --admin chris
`

Here's the terminal output with the error.

```
% Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  6354  100  6354    0     0  28115      0 --:--:-- --:--:-- --:--:-- 28115
Checking if TLJH is already installed...
TLJH already installed, upgrading...
Upgrading TLJH installer...
Setup tljh package
Starting TLJH installer...
Setting up admin users
Granting passwordless sudo to JupyterHub admins...
Setting up user environment...
Ran /opt/tljh/user/bin/python -m pip install -r /opt/tljh/hub/lib/python3.6/site-packages/tljh/requirements-base.txt with exit code 1
WARNING: The directory '/home/chris/.cache/pip/http' or its parent directory is not owned by the current user and the cache has been disabled. Please check the permissions and owner of that directory. If executing pip with sudo, you may want sudo's -H flag.
WARNING: The directory '/home/chris/.cache/pip' or its parent directory is not owned by the current user and caching wheels has been disabled. check the permissions and owner of that directory. If executing pip with sudo, you may want sudo's -H flag.
ERROR: Could not open requirements file: [Errno 2] No such file or directory: '/opt/tljh/hub/lib/python3.6/site-packages/tljh/requirements-base.txt'

Traceback (most recent call last):
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/opt/tljh/hub/lib/python3.6/site-packages/tljh/installer.py", line 508, in <module>
    main()
  File "/opt/tljh/hub/lib/python3.6/site-packages/tljh/installer.py", line 491, in main
    ensure_user_environment(args.user_requirements_txt_url)
  File "/opt/tljh/hub/lib/python3.6/site-packages/tljh/installer.py", line 277, in ensure_user_environment
[...]

Issue #503, opened 2020-02-04
Integration tests don't match the production installation workflow


manics:
For example https://github.com/jupyterhub/the-littlest-jupyterhub/pull/457 passed all CI tests, but causes an error in production: https://github.com/jupyterhub/the-littlest-jupyterhub/issues/502

yuvipanda:
I think @GeorgianaElena is right in https://github.com/jupyterhub/the-littlest-jupyterhub/pull/504#issuecomment-582280193.

However, our install code *also* git clones the repo, so I'm not sure why this was different.

GeorgianaElena:
> However, our install code also git clones the repo, so I'm not sure why this was different.

@yuvipanda, are you referring to the *git clone* that happens through [this](https://github.com/jupyterhub/the-littlest-jupyterhub/blob/master/bootstrap/bootstrap.py#L156) *pip install*?

I've read a bit through [these VCS support docs](https://pip.pypa.io/en/stable/reference/pip_install/#vcs-support), and it looks like when `--editable` is not used, the repo gets cloned somewhere in */tmp*. In the installer, *--editable* pip flag is added when `TLJH_BOOTSTRAP_DEV == 'yes'`.

For the integration test, we set *TLJH_BOOTSTRAP_DEV* to *yes* in the [Dockerfile](https://github.com/jupyterhub/the-littlest-jupyterhub/blob/master/integration-tests/Dockerfile#L24) and tljh gets cloned somewhere in the hub venv (I think). But when someone installs tljh, I don't think it's setting TLJH_BOOTSTRAP_DEV (I didn't), so tljh gets cloned in a tmp directory (and won't have access to the source code, only to the tljh package).

I believe this is what caused the *difference* between our tests and an actual deployment.

yuvipanda:
@GeorgianaElena ah, that makes sense! Should we turn that off in our CI? It's useful when developing locally...

GeorgianaElena:
I tried to turned it off, but nothing changed... :(
It appears that we're copying the source code (that comes from the git clone happening in the [checkout step](https://github.com/jupyterhub/the-littlest-jupyterhub/blob/master/.circleci/config.yml#L62)) into the container [here](https://github.com/jupyterhub/the-littlest-jupyterhub/blob/master/.circleci/integration-test.py#L88).

It think making the integration tests to work in a production-like manner, comes down to deploying TLJH in a Docker container. What do you think @yuvipanda?

PR #504, opened 2020-02-04
Fix missing reference to requirements-base.txt

GeorgianaElena:
Include `requirements-base.txt` in MANIFEST so it gets installed.
Fixes https://github.com/jupyterhub/the-littlest-jupyterhub/issues/502

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,3 @@
 include tljh/systemd-units/*
 include tljh/*.tpl
+include tljh/requirements-base.txt


yuvipanda:
Thanks, @GeorgianaElena!

betatim:
To catch errors with the manifest files we'd have to run our integration tests in/from a copy of the repo that was created by some kind of bundling tool right? I was surprised we didn't catch this in the integration tests and trying to understand why.

GeorgianaElena:
My guess is that it has something to do with the [checkout step](https://circleci.com/docs/2.0/configuration-reference/#checkout) in our circleci config.
Expand the [*checkout code* in our tests](https://circleci.com/gh/jupyterhub/the-littlest-jupyterhub/1845?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) to see what circleci is doing.

I believe because of this, even if we build tljh [here](https://github.com/jupyterhub/the-littlest-jupyterhub/blob/master/bootstrap/bootstrap.py#L156) without the requirements file lets say, it will still find the file because it has access to the source code. In a production scenario, this doesn't happen.

P.S. I'm not sure if I'm making any sense, this is just a hunch (it hard to understand how circleci works).

betatim:
Thanks for investigating. Debugging CI setups is no fun :-/ Looks like #503 is a good place to track and work on this issue.

Issue #239, opened 2018-11-23
Add upgrade CI


yuvipanda:
Our CI should be able to run upgrades from current master to any PR, and fail if the current suite of tests fail.

PR #511, opened 2020-02-17
Update tests

GeorgianaElena:
This PR makes the tljh installation in integration tests more simmilar to a production install.
This means:
- we are not copying anymore the source code into the test docker container
- we are only copying into the docker container the bootstrap script and the integration tests
- when we install tljh from git in the bootstrap script, we're not using the `--editable` flag anymore

Some small refactoring also happened:
 - Reuse venv setup command in unit + doc tests
 - Reuse docker executor in unit + doc tests

I've removed the `requirements-base.txt` from MANIFEST and made the bootstrap scrip install the branch with this PR to make sure this new version of the integration tests can catch this issue we had https://github.com/jupyterhub/the-littlest-jupyterhub/issues/502 and it did :confetti_ball: (circleci proof [here](https://circleci.com/gh/GeorgianaElena/the-littlest-jupyterhub/703)).

closes https://github.com/jupyterhub/the-littlest-jupyterhub/issues/503
 
closes https://github.com/jupyterhub/the-littlest-jupyterhub/issues/239


GeorgianaElena:
While trying to add an upgrade test (ref: https://github.com/jupyterhub/the-littlest-jupyterhub/issues/239), I realized that this PR is not ready yet as it always installs the master branch and not the PR branch as it should.

Have to find a way to make `tljh_repo_path` be the PR branch, so this is still a work in progress.

GeorgianaElena:
I believe I might have a review-able version :sparkles: 
Some things worth noting:
* CircleCi doesn't provide an environmental variable for the brach of the fork from where the PR was openend so I used this hack to take it:
```
PR_INFO_URL=https://api.github.com/repos/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME/pulls/$CIRCLE_PR_NUMBER
PR_BRANCH=`(curl -s $PR_INFO_URL | jq -r ".head.ref")`
```
* And this PR_BRANCH is used to form the TLJH_BOOTSTRAP_PIP_SPEC like this:
```
git+https://github.com/$CIRCLE_PROJECT_USERNAME/the-littlest-jupyterhub.git@$PR_BRANCH
```

betatim:
Ill take a look and start thinking :)

betatim:
Did some reading about the environment variables CircleCI gives us. I was thinking about how we could get the code from the PR/create the PIP spec without having to do the [gymnastics you mentioned](https://github.com/jupyterhub/the-littlest-jupyterhub/pull/511#issuecomment-588980506). It works but the fact that it requires so much acrobatics makes me think "woah, this will keep breaking". But this is just a hunch. Right now I don't have a good alternative idea.

So at the moment reading and thinking about how all this works in TLJH and the tests.

GeorgianaElena:
I'm not happy with that hack either, but I just didn't find a better way of doing it :cry:

.circleci/config.yml
```diff
@@ -1,33 +1,139 @@
...
+        steps:
+            - run:
+                name: Run basic tests
+                command: |
+                    PR_INFO_URL=https://api.github.com/repos/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME/pulls/$CIRCLE_PR_NUMBER
+                    PR_BRANCH=`(curl -s $PR_INFO_URL | jq -r ".head.ref")`
+                    if [ "$PR_BRANCH" == "null" ];then
+                        PR_BRANCH=$CIRCLE_BRANCH
+                    fi
+                    BOOTSTRAP_PIP_SPEC=git+https://github.com/$CIRCLE_PROJECT_USERNAME/the-littlest-jupyterhub.git@$PR_BRANCH
```
minrk: We shouldn't need to extract branche and repo info from the PR because the repo and commit being built should be enough:

```suggestion
BOOTSTRAP_PIP_SPEC=git+$CIRCLE_REPOSITORY_URL@$CIRCLE_SHA1
```

However, what might be even simpler is to use a *local path* to the current checkout, rather than referring to the repo. That can be another PR, though. If the current repo is mounted as a volume to the test container or already checked out there somehow, then we should be able to do

```suggestion
BOOTSTRAP_PIP_SPEC=file:///path/to/tljh-checkout
```

GeorgianaElena: Thanks for reviewing this @minrk.
If I understood things right, before this PR, we were actually using a local path to the current checkout:
* we were copying the current checkout code into /srv/src in the test container [here](https://github.com/jupyterhub/the-littlest-jupyterhub/blob/master/.circleci/integration-test.py#L88)
* TLJH_BOOSTRAP_PIP_SPEC was set to /srv/src in [the Dockerfile](https://github.com/jupyterhub/the-littlest-jupyterhub/blob/master/integration-tests/Dockerfile#L25)

But this approach, although simpler, led to [this error](https://github.com/jupyterhub/the-littlest-jupyterhub/issues/502) not being caught by the tests. This is why I tried keep in the test container only the integration tests and the bootstrap script and nothing else.

However, the suggestion to use the CIRCLE_SHA1 is really great and will help get rid of that ugly hack to extract the branch.

Thanks!

Issue #517, opened 2020-02-25
CondaUpgradeError


simonschnake:
I have installed TLJH on a google compute engine. I've followed the google cloud guide.
After a while I get the following error, when trying to install new packages:

```
CondaUpgradeError: This environment has previously been operated on by a conda version that's newer
than the conda currently being used. A newer version of conda is required.
  target environment location: /opt/tljh/user
  current conda version: 4.5.8
  minimum conda version: 4.8
```

Issue #332, opened 2019-05-19
Upgrade conda & hide the 'new version' nag during install


yuvipanda:
It's confusing a lot of people, and probably hiding other real errors. See https://github.com/jupyterhub/the-littlest-jupyterhub/issues/281 for example.

wqh17101:
i find that you check conda and traefik version which is a constant why not put the install script/package in the repo,and install from it . @yuvipanda

choldgraf:
Just a note that @poldrack ran into some issues with this today in a class that he was teaching. It sounds like the version of conda is out-of-date enough that it is starting to create bugs for users (I think there was an error where conda couldn't install anything because some kind of outdated version mismatch happened)

perhaps this is something @GeorgianaElena could help out with?

PR #519, opened 2020-02-26
Fix installer

GeorgianaElena:
There is a bug in the upgrade logic of the installer. On a fresh TLJH install, conda 4.8.1 is installed.
If the installer is run again, instead of using the existing conda 4.8.1, conda 4.5.8 is installed and used, producing a *CondaUpgradeError*. 

This should fix this issue.
Closes https://github.com/jupyterhub/the-littlest-jupyterhub/issues/517 
Closes https://github.com/jupyterhub/the-littlest-jupyterhub/issues/332


choldgraf:
amazing! thanks thanks thanks :-)

PR #521, opened 2020-02-26
run integration tests on latest Ubuntu 19.10

minrk:
reuses newly organized commands to make adding a second integration-test job with one different variable a little easier

Not sure if it's worth it (or possible) to make this a little less repetitive by declaring an integration_test *command* that runs all the other test commands, but that may not be worth it.

The ubuntu docker images use the `rolling` tag to mean the latest supported release (`latest` is latest LTS). We could use that instead of manually setting 19.10

- [ ] should we use `rolling` instead of `19.10` for the tests on latest ubuntu, or update ourselves every six months?
- [ ] should we use `latest` instead of `18.04` for the tests on LTS ubuntu, or update ourselves every two years or so?

includes #381


betatim:
The new and old integration tests are unhappy :(

GeorgianaElena:
I've rebased and all the tests are passing, now :sparkles:

> should we use rolling instead of 19.10 for the tests on latest ubuntu, or update ourselves every six months?

I vote for using rolling for the tests on latest ubuntu, to catch any breaking changes before people start reporting issues.

> should we use latest instead of 18.04 for the tests on LTS ubuntu, or update ourselves every two years or so?

I believe for this case, we should update ourselves  when enough time has passed (two years seems reasonable and it's not that frequent to become anoyning).

yuvipanda:
Now that 20.04 LTS is here, we should definitely test on that too...

consideRatio:
I suggest we merge #710 and close this, then open a new PR to resolve #714 that I created to represent the need for a test against 21.10 for example.

consideRatio:
I'll close this as resolved in #710, integration tests now run against 21.10, but not 19.10

PR #529, opened 2020-03-06
[WIP] Add --hub-requirements-txt-url to the installer

### LINKED ISSUE #525: Specify extra hub requirements
Similar to the [Installing python packages in the user environment](http://tljh.jupyter.org/en/latest/topic/customizing-installer.html#installing-python-packages-in-the-user-environment) section, it would be great to be able to install extra requirements for the hub as well.

This would be useful for example when using a spawner that is not part of this list:

https://github.com/jupyterhub/the-littlest-jupyterhub/blob/0656fbbb1495f7d2df5e3d990c783f2282d6551d/tljh/installer.py#L219-L228

The install command would then look something like this:

```bash
curl https://raw.githubusercontent.com/jupyterhub/the-littlest-jupyterhub/master/bootstrap/bootstrap.py \
 | sudo python3 - \
    --hub-requirements-txt-url https://raw.githubusercontent.com/path/to/hub/requirements.txt \
    --user-requirements-txt-url https://raw.githubusercontent.com/path/to/user/requirements.txt
```

For now I suppose this can be achieved by running the `pip install` command directly, but after the install process has completed, which then requires a restart of the hub.

The other option would be to make the requirements a plugin. But this requires extra steps and providing a folder with a `setup.py`.

jtpio:
Fixes #525.

Similar to the user requirements, this change lets specify extra hub requirements to the installer.

Although this could probably be achieved in a plugin by implementing the `tljh_extra_hub_pip_packages` hook, this flag makes it more convenient to install additional spawners and authenticators.

image

 - [ ] Add tests

### DIFF
--- a/docs/topic/customizing-installer.rst
+++ b/docs/topic/customizing-installer.rst
@@ -53,7 +53,7 @@ during install you would:
      | sudo python3 - \
        --admin admin-user1:password-user1 --admin admin-user2
 
-Installing python packages in the user environment
+Installing Python packages in the user environment
 ==================================================
 
 ``--user-requirements-txt-url <url-to-requirements.txt>`` installs packages specified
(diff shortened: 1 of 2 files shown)


jtpio:
Closing as this can be achieved with a plugin.

PR #531, opened 2020-03-09
Use PR username when no CircleCI project

GeorgianaElena:
1. When implementing https://github.com/jupyterhub/the-littlest-jupyterhub/pull/511, I didn't realize that when a TLJH fork doesn't have CircleCI setup there's no CIRCLE_PROJECT_USERNAME env var. Instead there is a  CIRCLE_PR_USERNAME that can be used.

Example:
* My TLJH fork has CircleCI setup, there are these env vars available:
project
* @minrk and @jtpio recent PRs don't have these env vars and instead have:
min
jtpio

2. Another way to solve this issue is installing from git refs (available since [pip 10](https://pip.pypa.io/en/stable/news/#id280)). However this would imply some gymnastics parsing the CIRCLE_PULL_REQUEST var.
circle_pr
like this:
```
index = CIRCLE_PULL_REQUEST.find("/pull/")
BOOTSTRAP_PIP_SPEC="git+" + CIRCLE_PULL_REQUEST[:index] + ".git@ref" + CIRCLE_PULL_REQUEST[index:] + "/head"
```

So that in the end to have BOOTSTRAP_PIP_SPEC be:
`git+https://github.com/jupyterhub/the-littlest-jupyterhub.git@ref/pull/527/head`

**What do folks think is the better way?**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,7 +42,11 @@ commands:
             - run:
                 name: Run basic tests
                 command: |
-                    BOOTSTRAP_PIP_SPEC=git+https://github.com/$CIRCLE_PROJECT_USERNAME/the-littlest-jupyterhub.git@$CIRCLE_SHA1
+                    if [ $CIRCLE_PR_USERNAME ]; then
+                        BOOTSTRAP_PIP_SPEC=git+https://github.com/$CIRCLE_PR_USERNAME/the-littlest-jupyterhub.git@$CIRCLE_SHA1
+                    else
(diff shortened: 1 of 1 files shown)


jtpio:
Thanks @GeorgianaElena for looking into this :+1: 

No strong opinion here, but it looks like the current change (option 1) works well and is quite explicit?

betatim:
Looks good to me, so merged it.

Do you know why the codecov part was missing?

PR #537, opened 2020-03-23
Update hub packages to the latest stable versions

jtpio:
This change updates the base hub packages to their latest stable version on PyPI.

### DIFF
--- a/tljh/installer.py
+++ b/tljh/installer.py
@@ -180,7 +180,7 @@ def ensure_jupyterlab_extensions():
         'labextension',
         'install'
     ] + extensions + install_options)
-    
+
     # Build all the lab extensions in one go using jupyter lab build command
     build_options = [
         '--minimize=False',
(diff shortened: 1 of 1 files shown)


jtpio:
Right, the `pydata-sphinx-theme` was released recently on PyPI and conda-forge.

I'll open a separate PR to fix it.

tljh/installer.py
```diff
@@ -217,14 +217,14 @@ def ensure_jupyterhub_package(prefix):
...
         'jupyterhub-systemdspawner==0.13',
-        'jupyterhub-firstuseauthenticator==0.13.0',
-        'jupyterhub-nativeauthenticator==0.0.4',
-        'jupyterhub-ldapauthenticator==1.2.2',
+        'jupyterhub-firstuseauthenticator==0.14.1',
+        'jupyterhub-nativeauthenticator==0.0.5',
+        'jupyterhub-ldapauthenticator==1.3.0',
         'jupyterhub-tmpauthenticator==0.6',
-        'oauthenticator==0.10.0',
+        'oauthenticator==0.11.0',
```
GeorgianaElena: As of OAuthenticator 0.11, there's no AWSCognito authenticator ([changelog ref](https://github.com/jupyterhub/oauthenticator/blob/master/docs/source/changelog.md#0110---2020-01-30)).

However, TLJH has a [section describing how to use AWSCognito](http://tljh.jupyter.org/en/latest/howto/auth/awscognito.html). If we update oauthenticator, that section should be removed or adapted so that it uses GenericOAuthenticator instead.

I can open up an issue after this PR gets merged and deal with it later on. What do you think @jtpio ? :eyes:

jtpio: Good catch :+1: 

We could also keep `oauthenticator==0.10.0` for now, and still open the issue?

jtpio:
> The docs build failure is unrelated and also happening in JH ([issue](https://github.com/jupyterhub/jupyterhub/issues/2993)). This [PR](https://github.com/jupyter/repo2docker/pull/864) solved it for repo2docker.

Thanks @GeorgianaElena for the heads-up! Just opened https://github.com/jupyterhub/the-littlest-jupyterhub/pull/538 to fix it.

tljh/installer.py
GeorgianaElena: Sure :sunny:

jtpio:
`oauthenticator` is now back at `0.10.0`.

PR #539, opened 2020-03-23
Pin oauthenticator to 0.10.0 for AWS Cognito

jtpio:
As mentioned by @GeorgianaElena in https://github.com/jupyterhub/the-littlest-jupyterhub/pull/537#discussion_r396497457, OAuthenticator 0.11 removed the AWS Cognito authenticator.

This change updates the docs to pin `oauthenticator` to `0.10.0` for now.


docs/howto/auth/awscognito.rst
```diff
@@ -71,8 +71,8 @@ able to configure the instance automatically, replace relevant config variables:
 Manual configuration to use the AWS Cognito Oauthenticator
 ============================================================
 
-Assuming tljh has already been installed, we need to make sure the oautheneticator module is at 0.10.0 and if not 
-do a pip install oauthenticator>=0.10.0
+Assuming tljh has already been installed, we need to make sure the oauthenticator module is at 0.10.0 and if not 
+do a pip install oauthenticator==0.10.0
```
GeorgianaElena: I think we should also mention that an AWSCongnito authenticator can still be achieved with oauthenticator 0.11, by configuring the GenericOAuthenticator class.

I tried documenting such a configuration in this PR https://github.com/jupyterhub/oauthenticator/pull/336, but I didn't test it however, so it might be incomplete. If it gets accepted, we could later link to the docs.

jtpio: Waiting for https://github.com/jupyterhub/oauthenticator/pull/336 to be merged sounds good :+1:

Just opened https://github.com/jupyterhub/the-littlest-jupyterhub/issues/540 to track this.

yuvipanda:
What do we do with this now?

yuvipanda:
Thinking about this some more, I think the way to do this is to create documentation on using the Generic OAuthenticator, and have subsection for AWS Cognito. What do you think of this direction, @jtpio?

jtpio:
Sounds good.

I won't be able to look into it right now, so converting to a draft in case someone would like to pick it up.

PR #553, opened 2020-04-22
Fix failing unit test

GeorgianaElena:
The unit test were failing because of the latest grpcio version 1.28.0, that switches to manylinux2010 (see [this announcement](https://groups.google.com/forum/#!msg/grpc-io/VdoTPMT5WwU/fMmXxFiBAwAJ) for details) (`grpcio` is a requirement of the `etcd3` python client used by traefik-proxy).

I tried a clean install of TLJH on DO and everything went ok, so referring to the info in the announcement I guess it has something to do with the Ubuntu Docker image and the libs there.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,6 +25,14 @@ commands:
                 command: |
                     apt-get update --yes && apt-get install --yes python3 python3-venv git make
 
+    upgrade_pip:
+        description: Upgrade the pip version to 20.0.*
+        steps:
+            - run:
+                name: upgrade pip
+                command: |
+                    python3 -m pip install -U pip==20.0.*
+
     build_systemd_image:
         steps:
             - run:
@@ -116,6 +124,9 @@ jobs:
                     - v1-dependencies-py3.6-
 
             - setup_venv
+
+            - upgrade_pip
+
             - run:
                 name: install dependencies
                 command: |
@@ -211,6 +222,7 @@ jobs:
                 key: v1-dependencies-py3.6-sphinx
 
             - setup_venv
+
             - run:
                 name: install dependencies
                 command: |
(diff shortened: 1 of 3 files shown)


manics:
It's because the default version of pip on Ubuntu 18.04 is too old:
```
$ docker run -it --rm ubuntu:18.04

root@68cd6b2855c2:/# apt-get update --yes && apt-get install --yes python3 python3-venv
root@68cd6b2855c2:/# python3 -m venv /srv/venv
root@68cd6b2855c2:/# /srv/venv/bin/pip --version
pip 9.0.1 from /srv/venv/lib/python3.6/site-packages (python 3.6)
```
The other option is to upgrade pip inside the venv
```
root@68cd6b2855c2:/# /srv/venv/bin/pip install -U pip
root@68cd6b2855c2:/# /srv/venv/bin/pip --version
pip 20.0.2 from /srv/venv/lib/python3.6/site-packages/pip (python 3.6)
```
I don't know what's best.

GeorgianaElena:
That's what was my first guess too. But TLJH got installed successfully on Digital Ocean using the same old pip:
tljh-do

I think this line in their announcement might explain why it works:
_Once the transition happens, if the version of pip is lower than 19, it might result in downloading a source package and building it from source instead of downloading pre-built package._

They also mention the need of `glibc` being at least 2.12 and `libstdc++` at least 6.0.13, but versions inside the docker container seem ok:
tljh-docker

So I'm not really sure what's the difference with the Docker image that causes this.

yuvipanda:
Thank you so much for investigating and getting to the root cause of this, @GeorgianaElena!

I think upgrading pip is a good idea anyway - although I like to pin it rather than just do a `--upgrade pip`. How does that sound? That's probably a better longer term fix than pinning grpcio, although that also works if that's the easier thing to do.

GeorgianaElena:
@yuvipanda, upgrading pip is definitely a good idea.
I just added a new commit that:
* upgrades pip inside the hub env
* upgrades pip inside the unit test docker container

What do you think?

yuvipanda:
Awesome! This looks great to me. Thank you for fixing the tests!

PR #556, opened 2020-04-23
Update the cull_idle_servers script

### LINKED ISSUE #547: Should the idle culler also cull named servers?
Testing a TLJH deployment with:

- named servers enabled (2 allowed per user)
- the idle culler service enabled

It looks like the default server is correctly culled after it has been idle for more than the number specified in [`timeout`](http://tljh.jupyter.org/en/latest/topic/idle-culler.html#idle-timeout), 

However the named servers are still shown as running in the list of servers and can be accessed normally, and they don't appear to be culled.

Should the idle culler script also take into account the named servers?

jtpio:
Update the `cull_idle_servers.py` script to the latest from JupyterHub: https://github.com/jupyterhub/jupyterhub/blob/8d505548494ad629cc9501f084c6ec4d7646cf98/examples/cull-idle/cull_idle_servers.py

This should fix #547.

- [ ] Add tests

### DIFF
--- a/tljh/cull_idle_servers.py
+++ b/tljh/cull_idle_servers.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 """script to monitor and cull idle single-user servers
 
-Imported from https://github.com/jupyterhub/jupyterhub/blob/6b1046697/examples/cull-idle/cull_idle_servers.py
+Imported from https://github.com/jupyterhub/jupyterhub/blob/8d505548494ad629cc9501f084c6ec4d7646cf98/examples/cull-idle/cull_idle_servers.py
 
 Caveats:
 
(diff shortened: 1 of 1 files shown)


jtpio:
Tested on a VM after setting the timeout to 60s with `sudo tljh-config set services.cull.timeout 60`.

After starting a named server:

image

~60 seconds later:

image

jtpio:
Some of the changes are coming from the `black` formatting in `jupyterhub`.

Also should we add tests for culling named servers? (in this PR or a separate one)
There seems to be some in https://github.com/jupyterhub/the-littlest-jupyterhub/blob/master/integration-tests/test_hub.py

GeorgianaElena:
> Also should we add tests for culling named servers? (in this PR or a separate one)
> There seems to be some in https://github.com/jupyterhub/the-littlest-jupyterhub/blob/master/integration-tests/test_hub.py

I believe adding a test would be very useful. Although the PR is pretty big already, I think it's better to have them together (code and test) in this PR to make sure it works before merging. What do you think?
 
Thanks for working on this @jtpio.

jtpio:
That would be better indeed.

Taking a closer look, it seems that we would need to add support for named servers to `hubtraf` too: https://github.com/yuvipanda/hubtraf/blob/37de44df5a2110c18bae4fb426566fddf1c1a05c/hubtraf/user.py#L80

jtpio:
This will actually be replaced by https://github.com/jupyterhub/the-littlest-jupyterhub/pull/559.

yuvipanda:
@jtpio damn, I did not see this! so sorry :(

jtpio:
No problem, https://github.com/jupyterhub/the-littlest-jupyterhub/pull/559 is much better!

jtpio:
Fixed by #559.

PR #558, opened 2020-04-27
Add bootstrap pip spec to the integration test docs

jtpio:
Update the docs to run the integration tests locally.

The `bootstrap_pip_spec` parameter can be specified before the test files to avoid shifting the arguments, similar to the Circle CI script:

https://github.com/jupyterhub/the-littlest-jupyterhub/blob/01ba34857dd4e316d839034ae2b3cc400b929964/.circleci/config.yml#L59-L61

image

 - [x] ~Add tests~

### DIFF
--- a/docs/contributing/tests.rst
+++ b/docs/contributing/tests.rst
@@ -8,17 +8,17 @@ Unit and integration tests are a core part of TLJH, as important as
 the code & documentation. They help validate that the code works as
 we think it does, and continues to do so when changes occur. They
 also help communicate in precise terms what we expect our code
-to do. 
+to do.
 
 Integration tests
 =================
 
-TLJH is a *distribution* where the primary value is the many 
+TLJH is a *distribution* where the primary value is the many
 opinionated choices we have made on components to use and how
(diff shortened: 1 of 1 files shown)


jtpio:
As an alternative, we might also want to specify the full bootstrap pip spec instead of `""`.

jtpio:
> And if they just want the default one, they just run the tests as usual, no empty string needed.

That would be really nice. An empty string is not very explicit and after reading the updated docs again (from this change), it felt a bit odd. An optional parameter sounds good :+1: 

> Update: Opened #563 with this change

Great, thanks!

I can update this PR to the mention the optional `bootstrap_pip_spec` in the docs. Or we can also close it and do everything in #563, that also works.

GeorgianaElena:
@jtpio, I've just merged https://github.com/jupyterhub/the-littlest-jupyterhub/pull/563.
I'd be great if you could update this PR :rocket:

jtpio:
Perfect! 

Just updated the PR and rebased on the latest master.

PR #563, opened 2020-04-30
Make bootstrap_pip_spec test argument optional

GeorgianaElena:
This makes `bootstrap_pip_spec` an optional argument so that when running the integration tests locally, users won't need to add the empty string as an arg (for the default tljh bootsrap pip spec) or the entire pip spec.
 
-~[ ] Add / update documentation~
 -~[ ] Add tests~

### DIFF
--- a/.circleci/integration-test.py
+++ b/.circleci/integration-test.py
@@ -158,8 +158,8 @@ def main():
     run_test_parser = subparsers.add_parser('run-test')
     run_test_parser.add_argument('--installer-args', default='')
     run_test_parser.add_argument('--upgrade', action='store_true')
+    run_test_parser.add_argument('--bootstrap_pip_spec', nargs='?', default="", type=str)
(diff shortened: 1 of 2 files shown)


.circleci/integration-test.py
```diff
@@ -158,8 +158,8 @@ def main():
     run_test_parser = subparsers.add_parser('run-test')
     run_test_parser.add_argument('--installer-args', default='')
     run_test_parser.add_argument('--upgrade', action='store_true')
+    run_test_parser.add_argument('--bootstrap_pip_spec', nargs='?', default="", type=str)
```
jtpio, line 161: Wondering whether the parameter should be named `--bootstrap-pip-spec`, similar to `--installer-args`? (so they both follow the same style)

GeorgianaElena: Oh, right :confused: Yes, that would be better.

PR #570, opened 2020-05-08
Use sha256 sums for verifying miniconda download

yuvipanda:
miniconda downloader now provides sha256 hashes. Previously
it provided only md5 hashes. md5 is deprecated everywhere for
everything, so let's use this

 - [x] Add tests

### DIFF
--- a/tests/test_conda.py
+++ b/tests/test_conda.py
@@ -14,9 +14,10 @@ def prefix():
     Provide a temporary directory with a conda environment
     """
     miniconda_version = '4.7.10'
-    miniconda_installer_md5 = "1c945f2b3335c7b2b15130b1b2dc5cf4"
+    miniconda_installer_sha256 = "8a324adcc9eaf1c09e22a992bb6234d91a94146840ee6b11c114ecadafc68121"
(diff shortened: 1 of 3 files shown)


yuvipanda:
This also gives us the possibility of using [miniforge](https://github.com/conda-forge/miniforge/) if we want to. I'd love for us to move to that eventually, as it is more of a community effort than miniconda.

yuvipanda:
This looks like an out of memory error again, but with conda install this time.

GeorgianaElena:
@yuvipanda, this time the error doesn't go away after a few test reruns  :confused:. 
So, I think we should bump the memory limit again. WDYT?

GeorgianaElena:
@yuvipanda, I've ended up increasing the [memory to 1.15G](https://github.com/jupyterhub/the-littlest-jupyterhub/pull/580) and rebased this PR to make everything green again.

I'm going to hit merge. Thanks @yuvipanda :tada: :heart:

yuvipanda:
Sorry I missed this, @GeorgianaElena! Thanks for taking care of it! <3

PR #578, opened 2020-05-22
Allow installation on unsupported OS

coreyralph:
TLJH successfully installs and runs on Debian Buster. This patch updates bootstrap.py to allow it.

Tested on Debian Buster and Ubuntu Bionic.

### DIFF
--- a/bootstrap/bootstrap.py
+++ b/bootstrap/bootstrap.py
@@ -74,7 +74,7 @@ def validate_host():
     if distro != 'ubuntu':
         print('The Littlest JupyterHub currently supports Ubuntu Linux only')
         sys.exit(1)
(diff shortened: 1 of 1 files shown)


willirath:
Not sure it is a good idea to explicitly support more than one OS as it increases load in maintenance and documentation.

To make it easier to run the bootstrap script on a non supported OS, I'd rather add a CLI flag that skips the OS check.

coreyralph:
> I'd rather add a CLI flag that skips the OS check.

Makes sense. I've updated the PR. Should this flag be documented?

> Wouldn't that also increase the maintenance, as there would be more CLI flags to maintain?

I think maintenance will be minimal. This flag is only used by `bootstrap.py` and not passed to the installer. The logic is quite simple and self-contained.

Issue #543, opened 2020-03-30
Using a custom traefik toml file or template


hMED22:
I'm trying with @jhadjar to change the proxy config in `/opt/tljh/state/traefik.toml` but it is overwritten by [traefik.toml.tpl ]( https://github.com/jupyterhub/the-littlest-jupyterhub/blob/master/tljh/traefik.toml.tpl ) every time the proxy is reloaded, is there any way to set a custom file or template to be used instead?

mskblackbelt:
Seconded. I'm currently restoring from a backup file each time, but it doesn't seem like the process should be that difficult. Alternately, is there a (supported) way to place some configuration values in `jupyterhub_config.d` to have some data appended to the existing TOML file?

jhadjar:
Adding more context to @hMED22's issue. The problem lies in https://github.com/jupyterhub/the-littlest-jupyterhub/blob/master/tljh/traefik.py#L90

```python
def ensure_traefik_config(state_dir):
    """Render the traefik.toml config file"""
    config = load_config()
    config['traefik_api']['basic_auth'] = compute_basic_auth(
        config['traefik_api']['username'],
        config['traefik_api']['password'],
    )

    with open(os.path.join(os.path.dirname(__file__), "traefik.toml.tpl")) as f:
        template = Template(f.read())
```

The problematic line gives no way to pass the template programmatically:

```python
    with open(os.path.join(os.path.dirname(__file__), "traefik.toml.tpl")) as f:
```

GeorgianaElena:
Indeed, `traefik.toml.tpl` is meant to be modified through `tljh-config` (to the extent of an option being configurable at all).

Can you please explain a bit what's your use-case and how would modifying `traefik.toml.tpl` help?

Some of the config in traefik.toml has to be there for TLJH and traefik-proxy to be functional. However, I think it is possible to add an *escape hatch* to extend the `traefik.toml` template, similarly with how we do for the `juptyerhub_config.py.`
WDYT about this @yuvipanda?

mskblackbelt:
I'm trying to set the `rules.toml` file to look like so:

```toml
[frontends]
  [frontends.default]
  backend = "backend__2F"
  passHostHeader = true
  priority = 80
    [frontends.default.routes.test]
    rule = "Path:/;AddPrefix:/hub"
    data = "{\"hub\": true}"

  [frontends.frontend__2F]
  backend = "backend__2F"
  passHostHeader = true
  priority = 90
[...]

PR #582, opened 2020-06-02
Allow extending traefik config

GeorgianaElena:
This allows extending traefik's config by adding `*.toml` snippets of traefik config in `traefik_config.d` directory. This is similar with how unsupported JupyterHub config is plugged into TLJH.

Unfortunately, since Traefik < 2.0 doesn't allow `yaml` configs I had to stick to `toml`. 

 - [x] Add tests

Closes https://github.com/jupyterhub/the-littlest-jupyterhub/issues/543

### DIFF
--- a/docs/topic/escape-hatch.rst
+++ b/docs/topic/escape-hatch.rst
@@ -63,3 +63,5 @@ proxy for the new configuration to take effect:
 .. code-block:: bash
 
 	sudo tljh-config reload proxy
+
+.. warning:: This instructions might change when TLJH will switch to Traefik > 2.0
--- a/docs/topic/escape-hatch.rst
+++ b/docs/topic/escape-hatch.rst
@@ -12,7 +12,7 @@ The two main TLJH components are **JupyterHub** and **Traefik**.
 
 These files are created by TLJH during installation and can be edited by the
 user only through ``tljh-config``. Any direct modification to these files,
-will have no effect.
+is unsupported, and will cause hard to debug issues.
 
 But because sometimes TLJH needs to be customized in ways that are not officially
 supported, an escape hatch has been introduced to allow easily extending the
--- a/docs/topic/escape-hatch.rst
+++ b/docs/topic/escape-hatch.rst
@@ -11,7 +11,7 @@ The two main TLJH components are **JupyterHub** and **Traefik**.
 * Traefik takes its configuration from the ``traefik.toml`` file.
 
 These files are created by TLJH during installation and can be edited by the
-user only through ``tljh-config``. Any direct modification to these files,
+user only through ``tljh-config``. Any direct modification to these files
 is unsupported, and will cause hard to debug issues.
 
 But because sometimes TLJH needs to be customized in ways that are not officially
--- a/docs/topic/escape-hatch.rst
+++ b/docs/topic/escape-hatch.rst
@@ -11,8 +11,8 @@ The two main TLJH components are **JupyterHub** and **Traefik**.
 * Traefik takes its configuration from the ``traefik.toml`` file.
 
 These files are created by TLJH during installation and can be edited by the
-user only through ``tljh-config``, so any direct modification to these files,
-will be overwritten by TLJH when the components are reloaded.
+user only through ``tljh-config``. Any direct modification to these files,
+will have no effect.
 
 But because sometimes TLJH needs to be customized in ways that are not officially
 supported, an escape hatch has been introduced to allow easily extending the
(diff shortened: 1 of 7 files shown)


tests/test_traefik.py
```diff
@@ -124,3 +124,33 @@ def test_manual_ssl_config(tljh_dir):
             "whiteList": {"sourceRange": ["127.0.0.1"]}
         },
     }
+
+def test_extra_config(tmpdir, tljh_dir):
+    extra_config_dir = os.path.join(tljh_dir, config.CONFIG_DIR, "traefik_config.d")
+
+    state_dir = tmpdir.mkdir("state")
+
+    extra_config = {
```
yuvipanda, line 133: I would love for us to add a small integration test here as well - to make sure that traefik actually loads the extra config and applies it. For example, with `"dashboard": True`, we could maybe hit an endpoint to see if the dashboard is running? Or something like that... 

What do you think?

GeorgianaElena: Sure! This is a great idea :sparkles: 
Thanks Yuvi :heart:

GeorgianaElena:
@yuvipanda, I just added the docs and an integration test.

The flow changed a bit since your last review because it turns out that doing
`toml.load([file1, file2])` merges the two files into a dict using `dict.update` and that overwrites some keys when the dicts are nested (which is often the case for traefik). So, I ended up using our recursive `_merge_dictionaries` function to merge the defaults with the extra configs.

docs/topic/escape-hatch.rst
```diff
@@ -1,17 +1,67 @@
...
+=============================
+Custom configuration snippets
+=============================
+
+The two main TLJH components are **JupyterHub** and **Traefik**.
+
+* JupyterHub takes its configuration from the ``jupyterhub_config.py`` file.
+* Traefik takes its configuration from the ``traefik.toml`` file.
+
+These files are created by TLJH during installation and can be edited by the
```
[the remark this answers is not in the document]
GeorgianaElena: Oh right! Thanks, Yuvi!

integration-tests/test_proxy.py
```diff
@@ -53,14 +55,64 @@ def test_manual_https(preserve_config):
...
+    }
+
+    success = False
+    for i in range(5):
+        time.sleep(i)
+        try:
+            with pytest.raises(HTTPClientError, match="HTTP 401: Unauthorized"):
+                # The default dashboard entrypoint requires authentication, so it should fail
+                req = HTTPRequest("http://127.0.0.1:8099/dashboard/", method="GET")
+                HTTPClient().fetch(req)
```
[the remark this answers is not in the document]
GeorgianaElena: This shouldn't return 200 because I'm not passing the credentials through the request, but it's expecting a 401:
 ```
with pytest.raises(HTTPClientError, match="HTTP 401: Unauthorized")
```
The boolean is to check that we exited the loop because we got the exception.

I could pass the credentials and verify with a 200 code instead? What do you think?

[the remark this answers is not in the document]
Comment (yuvipanda) [reply integration-tests/test_proxy.py:91]: Aaaah, I missed the raises! This looks great, sorry about that!

File: docs/topic/escape-hatch.rst [hunk 4]
```diff
@@ -11,8 +11,8 @@ The two main TLJH components are **JupyterHub** and **Traefik**.
 * Traefik takes its configuration from the ``traefik.toml`` file.
 
 These files are created by TLJH during installation and can be edited by the
-user only through ``tljh-config``, so any direct modification to these files,
-will be overwritten by TLJH when the components are reloaded.
+user only through ``tljh-config``. Any direct modification to these files,
+will have no effect.
```
yuvipanda, line 15:
```suggestion
is unsupported, and will cause hard to debug issues.
```

docs/topic/escape-hatch.rst
```diff
@@ -11,8 +11,8 @@ The two main TLJH components are **JupyterHub** and **Traefik**.
 * Traefik takes its configuration from the ``traefik.toml`` file.
 
 These files are created by TLJH during installation and can be edited by the
-user only through ``tljh-config``, so any direct modification to these files,
-will be overwritten by TLJH when the components are reloaded.
+user only through ``tljh-config``. Any direct modification to these files,
```
yuvipanda, line 14:
```suggestion
user only through ``tljh-config``. Any direct modification to these files
```

GeorgianaElena:
Thanks @yuvipanda :sunflower:

yuvipanda:
\o/ Thanks for working on this, @GeorgianaElena!

PR #593, opened 2020-06-29
add -L option to curl to follow redirect

LTangaF:
to new bootstrap location

 - [ ] Add tests

### DIFF
--- a/.circleci/integration-test.py
+++ b/.circleci/integration-test.py
@@ -100,7 +100,7 @@ def run_test(image_name, test_name, bootstrap_pip_spec, test_files, upgrade, ins
     if upgrade:
         run_container_command(
             test_name,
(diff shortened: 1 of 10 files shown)


LTangaF:
This is a minor documentation fix. Not sure how testable that is from an automated perspective. I tested it in AWS, though.

LTangaF:
I didn't realize there was a test for the documentation changes. It looks like it came up green here without it, which is interesting.

I'll work on those other doc changes.

GeorgianaElena:
> I didn't realize there was a test for the documentation changes. It looks like it came up green here without it, which is interesting.

There isn't any test for doc changes that would error from this. I think the red from my `grep` command output was misleading. Sorry for that.

> I'll work on those other doc changes.

Thanks a lot for working on this :heart:

LTangaF:
I updated the integration test as well.

PR #599, opened 2020-07-21
ensure_server is now ensure_server_simulate

GeorgianaElena:
Update the tests to keep up with the great things happening in [hubtraf](https://github.com/yuvipanda/hubtraf/tree/master/hubtraf) (:heart:  @yuvipanda )


yuvipanda:
Wow I forgot that hubtraf has users!!! :D <3

PR #602, opened 2020-07-29
Bump systemdspawner

yuvipanda:
New version released, with https://github.com/jupyterhub/systemdspawner/pull/53

 - [ ] Add tests


GeorgianaElena:
The upgrade test failed because jupyterlab failed to build (probably because memory issues again :( )
I'm wondering if jupyterlab 2.* requires more memory to build. I merged the PR with the jupyterlab upgrade yesterday.

Anyway, I believe we should merge this since the failures are unrelated.

Issue #70, opened 2018-07-22
Serve a temporary "TLJH is building" page while TLJH is building


choldgraf:
Would it be straightforward for TLJH to first set up a really short-and-sweet "please be patient while your Littlest JupyterHub is built" page? This would help users disambiguate "I just need to wait a bit longer for my jupyterhub to set up" and "something went wrong with the startup script and I need to debug"

PR #605, opened 2020-08-03
Temporary page while tljh is building

GeorgianaElena:
This shows a temporary page while tljh is building, that looks like this:
tljh-is-building-page

A few things to consider:
* There will be a "down" period between when this page stops showing and JupyterHub starts, because we have to free up the port for traefik.
* The `/logs` page has to be manually refreshed to update the logs
* The main page gets refreshed every 30s so that when JH starts, the login page will be shown instead

TODO
 - [x] Add / update tests

Closes https://github.com/jupyterhub/the-littlest-jupyterhub/issues/70


bootstrap/bootstrap.py
```diff
@@ -90,7 +93,38 @@ def validate_host():
...
         sys.exit(1)
 
+class LoaderPageRequestHandler(SimpleHTTPRequestHandler):
+    def do_GET(self):
+        if self.path == "/logs":
+            with open("/opt/tljh/installer.log", "rb") as log_file:
+                content = log_file.read()
+                self.wfile.write(content)
+        else:
+            self.path = '/index.html'
```
yuvipanda, line 103: For routing, I am thinking:

1. if it's '/logs', do what we're doing now
2. if it's for '/', redirect to '/index.html'
3. If it's for '/index.html', serve the file
4. For anything else, return a '403 Forbidden' or '404 Not Found'

We want to only serve the files we want to expose to the public, since otherwise we might accidentally expose secrets to the whole world. A routing approach that explicitly lists routes allowed helps with this.

GeorgianaElena:
Thanks for the feedback @yuvipanda :)
I addressed all of the suggestions, except for the password one.

> I also wonder if this should be behind password protection, in case the logs have sensitive information?

There shouldn't be anything sensitive in the logs, sice we're only logging generic messages at each step. I think there might be some in case errors occur, but those go only to `stderr` and are not stored into `installer.log`.

But if we decide to password protect the temporary page, maybe one way would be to use the user and password of the first `tljh` admin user, that are passed to the `bootstrap` script.

What do you think?

bootstrap/index.html
```diff
@@ -0,0 +1,54 @@
+<!DOCTYPE html>
+<head>
+  <body>
+    <meta http-equiv="refresh" content="30" >
+    <meta http-equiv="content-type" content="text/html; charset=utf-8">
+    <meta name="viewport" content="width=device-width">
+    <img style="width: 150px; height: auto;" src="https://raw.githubusercontent.com/jupyterhub/the-littlest-jupyterhub/master/docs/images/logo/logo.png">
+    <div class="loader center"></div>
+    <div class="center" style="font-size: 30px; font-weight: bold; color: grey; text-align:center;">Please wait while your TLJH is building...</div>
```
yuvipanda, line 9: Let's put all this under the `<style>` tag and avoid inline styles?

bootstrap/index.html
```diff
@@ -0,0 +1,54 @@
+<!DOCTYPE html>
+<head>
+  <body>
```
yuvipanda, line 3: `<body>` should be outside the `<head>`

yuvipanda: We should also add a `<title>` tag

tljh/installer.py
```diff
@@ -485,6 +486,10 @@ def main():
         nargs='*',
         help='Plugin pip-specs to install'
     )
+    argparser.add_argument(
+        '--temporary-page',
```
yuvipanda, line 490: If this is taking the `pid` of the process showing the temporary page, we should rename this to something like `--progress-page-server-pid` or similar? Could also add `type=int` to enforce it's an integer.

tljh/installer.py
```diff
@@ -499,6 +504,17 @@ def main():
     ensure_node()
     ensure_jupyterhub_package(HUB_ENV_PREFIX)
     ensure_jupyterlab_extensions()
+
+    # Stop the http server with the loading page before traefik starts
+    if args.temporary_page:
+        try:
+            os.kill(int(args.temporary_page), signal.SIGINT)
+            # Remove the pid file and the temporary html page
+            os.remove('/index.html')
```
yuvipanda, line 513: We should put everything under `/var/run` or someplace, and not under `/`. This helps keep `/` clean!

bootstrap/bootstrap.py
```diff
@@ -90,7 +93,47 @@ def validate_host():
...
+
+def serve_forever(server):
+    try:
+        server.serve_forever()
+    except KeyboardInterrupt:
+        pass
+
 def main():
+    temp_page_flag = "--temporary-page"
+    if temp_page_flag in sys.argv:
```
yuvipanda, line 119: Since `argparse` is in the standard library, what do you think of using that here? Would be cleaner than directly looking in `sys.argv`. However, we do pass all params through to installer, so if it is difficult we should add a comment here on why this isn't using argparse, and let it go :)

bootstrap/bootstrap.py
```diff
@@ -90,7 +93,47 @@ def validate_host():
...
+        # Serve the loading page until TLJH builds
+        index_url="https://raw.githubusercontent.com/GeorgianaElena/the-littlest-jupyterhub/in-progress-page/bootstrap/index.html"
```
yuvipanda, line 121: Hmm, so this means the contents of `index.html` will evolve independently of bootstrap.py. What do you think about embedding the HTML file directly in bootstrap.py?

bootstrap/bootstrap.py
```diff
@@ -90,7 +93,47 @@ def validate_host():
...
+        urllib.request.urlretrieve(index_url, "index.html")
+        urllib.request.urlretrieve(favicon_url, "favicon.ico")
+
+        # If the bootstrap is run to upgrade TLJH, then this will raise an "Address already in use" error
+        try:
+            loading_page_server = HTTPServer(("", 80), LoaderPageRequestHandler)
+            p = multiprocessing.Process(target=serve_forever, args=(loading_page_server,))
+            p.start()
+            # Put the pid after the temp page flag to be passed to the istaller
+            sys.argv.insert(sys.argv.index(temp_page_flag)+1, str(p.pid))
```
yuvipanda, line 132: We should probably make a copy of sys.argv earlier in `main()` and manipulate that, rather than modify sys.argv directly.

yuvipanda:
> But if we decide to password protect the temporary page, maybe one way would be to use the user and password of the first `tljh` admin user, that are passed to the `bootstrap` script.

Yep, that's the way to do it I think! But we can add that as a separate PR, not necessarily part of this one.

yuvipanda:
what kinda tests do you think we should add to this?

bootstrap/bootstrap.py
GeorgianaElena: I've added a comment about not using `argparse` :sweat_smile:

The reasoning for that was that if using `argparse` here for the temporary page flag, all the other flags should be defined here also (I got complaints about not knowing the other args).  And I think apart from this temp page all of the flags are installer-specific. WDYT?

GeorgianaElena:
I added a test in `test_bootrap.py` that checks if there is something at `/index.html` while tljh is building and that the same path returns a 404 once jupyterhub starts.

Do you think this is a good enough test @yuvipanda?

integration-tests/test_bootstrap.py
```diff
@@ -1,28 +1,36 @@
...
         '/bin/bash', '-c', 'sleep 1000s'
     ])
     # Install python3 inside the ubuntu container
     # There is no trusted Ubuntu+Python3 container we can use
     subprocess.check_output([
         'docker', 'exec', container_name, 'apt-get', 'update'
     ])
+
+    if flags:
+        pkgs = ['python3', 'systemd', 'git']
```
yuvipanda, line 28: Curious - why do we need git & systemd only if flags is set?

tljh/installer.py
```diff
@@ -499,6 +505,14 @@ def main():
     ensure_node()
     ensure_jupyterhub_package(HUB_ENV_PREFIX)
     ensure_jupyterlab_extensions()
+
+    # Stop the http server with the loading page before traefik starts
+    if args.progress_page_server_pid:
+        try:
+            os.kill(args.progress_page_server_pid, signal.SIGINT)
+        except Exception as e:
+            pass
```
yuvipanda, line 514: Let's log an error message when this fails?

yuvipanda:
Things look good to me, @GeorgianaElena! I left some minor questions. Once we get the tests to pass this can be merged! \o/

GeorgianaElena:
Tests are now passing @yuvipanda, if you want to take another look :tada:

docs/topic/customizing-installer.rst
```diff
@@ -17,6 +17,35 @@ This page documents the various options you can pass as commandline parameters t
...
+Serving a temporary "TLJH is building" page
+===========================================
+``--show-progress-page`` serves a temporary "TLJH is building" progress page while TLJH is building.
+
+.. image:: ../images/tljh-is-building-page.gif
+  :alt: Temporary progress page while TLJH is building
+
+* The page will be accessible at ``http://<tljh-public-ip>/index.html`` in your browser.
+  When TLJH installation is complete, the progress page page will stop and you will be able
+  to access TLJH as usually at ``http://<tljh-public-ip>/index.html``.
```
yuvipanda, line 29:
```suggestion
  to access TLJH as usually at ``http://<tljh-public-ip>``.
```

yuvipanda:
The upgrade test is failing thanks to jupyterlab again, so I'm just going to merge this.

AWESOME WORK, @GeorgianaElena! \o/

PR #621, opened 2020-10-19
Fix HTML of bootstrap

richardbrinkman:
The <html> tag was not correctly closed with a </html> tag.


GeorgianaElena:
Thanks @richardbrinkman :tada:

PR #632, opened 2020-11-14
how to call sudo with changed path

namin:


GeorgianaElena (approved):
Thanks @namin! I noticed that on my Ubuntu it works both with and without `env`, but it might be different for other distributions (looking at the  `sudo` man page for CentOs, it looks like it explicitly needs the `env` part). So I'm going to merge this :rocket: Thank you!

PR #633, opened 2020-11-16
proposed changes for issue #619

### LINKED ISSUE #619: Move implementation of class UserCreatingSpawner in a separate module
### Proposed change

I want a TLJH setup in which every user is added automatically to a specific secondary group. This is currently not covered through the `users.extra_user_groups` configuration (which requires to specify the user names), nor through any Linux configuration I am aware of (e.g., in `/etc/login.defs`) or any other way that I was able to figure out. 

However, it is very easy to implement by extending the functionality of class `UserCreatingSpawner`. What I would like to do (in a JupyterHub config file) is the following:
```
from tljh.jupyterhub_config import UserCreatingSpawner
from tljh import user
from tljh.normalize import generate_system_username

class CustomUserCreatingSpawner(UserCreatingSpawner):
    def start(self):
        super_start = super().start()
        system_username = generate_system_username('jupyter-' + self.user.name)
        user.ensure_user_group(system_username, 'testgroup')
        return super_start

c.JupyterHub.spawner_class = CustomUserCreatingSpawner
```
Unfortunately, when I run `tljh-config reload` the import statement in the first line causes the reload to hang. By moving the implementation of class `UserCreatingSpawner` into a separate module, e.g., `tljh/user_creating_spawner.py`, this issues can be solved.

### Alternative options

Alternatively, there could be a configuration option to add all users to a secondary user group (unlike `users.extra_user_groups` which requires to specify the user names). But this requires a bigger effort to implement than my proposed solution.

I can imagine that there are also other reasons to modify class `UserCreatingSpawner`. My proposed change would benefit also these cases.

Apart from that, it would be a *cleaner* solution in general, separating class implementations and configuration scripts.

Class UserCreatingSpawner is currently defined in `tljh/jupyterhub_config.py`. I propose to put it into a separate module, e.g., `tljh/user_creating_spawner.py`.

I would be happy to forge the repo, implement this minor change and make an official merge request.

ewidl:
Fixes #619 

This is only a minor change for the project, which has no direct impact on the functionality. However, it would make my life a bit easier, resulting in less code duplication in my own TLJH setup (see description of issue #619). Also, it would be a *cleaner* solution in general, separating class implementations and configuration scripts.

### DIFF
--- a/tljh/jupyterhub_config.py
+++ b/tljh/jupyterhub_config.py
@@ -6,42 +6,11 @@ from glob import glob
 import os
 
 from systemdspawner import SystemdSpawner
-from tljh import configurer, user
+from tljh import configurer
 from tljh.config import INSTALL_PREFIX, USER_ENV_PREFIX, CONFIG_DIR
-from tljh.normalize import generate_system_username
 from tljh.utils import get_plugin_manager
+from tljh.user_creating_spawner import UserCreatingSpawner
 from jupyterhub_traefik_proxy import TraefikTomlProxy
-
-from traitlets import Dict, Unicode, List
-
-class UserCreatingSpawner(SystemdSpawner):
-    """
-    SystemdSpawner with user creation on spawn.
-
-    FIXME: Remove this somehow?
-    """
-    user_groups = Dict(key_trait=Unicode(), value_trait=List(Unicode()), config=True)
-
-    def start(self):
-        """
-        Perform system user activities before starting server
-        """
-        # FIXME: Move this elsewhere? Into the Authenticator?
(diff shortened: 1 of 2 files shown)


meeseeksmachine:
This pull request has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/move-implementation-of-class-usercreatingspawner-in-a-separate-module/6827/1

GeorgianaElena (approved):
Thank you @ewidl! I don't see any harm in having `UserCreatingSpawner` in its own module, especially since it helps with your use-case, so I'm going to merge this :rocket:

ewidl:
Thank you!

PR #635, opened 2020-11-24
[WIP] Upgrade to ubuntu 20

GeorgianaElena:
- [x] Check for Ubuntu 20
 - [x] Upgrade conda version to 4.8.5
 - [x] Use miniforge
 - [ ] Add tests
 - [ ] Todo: see what else needs to be removed/changed

### DIFF
--- a/bootstrap/bootstrap.py
+++ b/bootstrap/bootstrap.py
@@ -131,18 +131,15 @@ def run_subprocess(cmd, *args, **kwargs):
     logger = logging.getLogger('tljh')
     proc = subprocess.run(cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, *args, **kwargs)
     printable_command = ' '.join(cmd)
-    if proc.returncode != 0:
+    code=proc.returncode
+    if code != 0:
         # Our process failed! Show output to the user
-        logger.error('Ran {command} with exit code {code}'.format(
-            command=printable_command, code=proc.returncode
-        ))
+        logger.error(f"Ran {printable_command} with exit code {code}")
         logger.error(proc.stdout.decode())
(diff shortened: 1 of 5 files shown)


bootstrap/bootstrap.py
```diff
@@ -8,7 +8,7 @@
     curl <script-url> | sudo python3 -
 
 Constraints:
-  - Entire script should be compatible with Python 3.6 (We run on Ubuntu 18.04+)
+  - Entire script should be compatible with Python 3.8 (We run on Ubuntu 20.04+)
   - Script should parse in Python 3.4 (since we exit with useful error message on Ubuntu 14.04+)
```
manics, line 12:
```suggestion
  - Script should parse in Python 3.6 (since we exit with useful error message on Ubuntu 18.04+)
```
Since you've added f-strings.

More changes will be need though, e.g.
https://github.com/jupyterhub/the-littlest-jupyterhub/blob/8ba0ca2a9efbe1958c864cd338c6b79f025dff87/integration-tests/test_bootstrap.py#L87-L88

GeorgianaElena: Thanks @manics!

consideRatio:
@GeorgianaElena this is a great maintenance PR! I'd love to help it get merged in any way I can.

I note the integration tests fail [here](https://github.com/jupyterhub/the-littlest-jupyterhub/pull/635/checks?check_run_id=3915174418)...

```python
    def test_progress_page():
        with concurrent.futures.ThreadPoolExecutor() as executor:
            installer = executor.submit(
                run_bootstrap, "progress-page", "ubuntu:20.04", True
            )
    
            # Check if progress page started
            started = verify_progress_page(expected_status_code=200, timeout=120)
>           assert started
```

Because there is a connection refused situation reaching localhost:80 apparently, I haven't investigated why.

PR #637, opened 2020-11-26
Doc of how users can change password

mauro3:
Doc of how users can change password for First Use Authenticator.


GeorgianaElena (approved):
Awesome! Thanks @mauro3 :tada:

Issue #642, opened 2020-12-18
tljh-config ContextualVersionConflict - chardet 4 vs jupyterhub-traefik-proxy


ixcat:
### Bug description

tljh-config failing to run after a fresh container rebuild of integration-tests/Dockerfile . Was working for me last install round.

as best I can tell - something in the tljh install is installing chardet 4, which conflicts with the pinned `aiohttp==3.6.2` in jupyterhub-traefik-proxy; tried to pip upgrade chardet/aiohttp, but this triggered errors. Not sure if it would be better to file against `jupyterhub-traefik-proxy` or here; filing here since this is where I encountered the error, and concievably the aiohttp pin is deliberate in `jupyterhub-traefik-proxy`

was able to resolve via:

```
. /opt/tljh/hub/bin/activate
pip uninstall chardet
pip install chardet==3.0.4
```

#### Expected behaviour

`tljh-config` should run

#### Actual behaviour

```
root@b65e93e47512:/srv/src# tljh-config set auth.GenericOAuthenticator.client_id 'client1'
Traceback (most recent call last):
  File "/opt/tljh/hub/lib/python3.6/site-packages/pkg_resources/__init__.py", line 574, in _build_master
    ws.require(__requires__)
  File "/opt/tljh/hub/lib/python3.6/site-packages/pkg_resources/__init__.py", line 892, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/opt/tljh/hub/lib/python3.6/site-packages/pkg_resources/__init__.py", line 783, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (chardet 4.0.0 (/opt/tljh/hub/lib/python3.6/site-packages), Requirement.parse('chardet<4.0,>=2.0'), {'aiohttp'})

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/bin/tljh-config", line 6, in <module>
    from pkg_resources import load_entry_point
  File "/opt/tljh/hub/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3088, in <module>
    @_call_aside
  File "/opt/tljh/hub/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3072, in _call_aside
    f(*args, **kwargs)
  File "/opt/tljh/hub/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3101, in _initialize_master_working_set
[...]

PR #643, opened 2020-12-23
Pin chardet

GeorgianaElena:
* Pin `chardet` in the hub environment to a version that doesn't conflict with traefik-proxy's `aiohttp` requirement until we figure out what is going on.
* Bump the `jupyterhub` version in the hub env to the same version in the user env

Reported in https://github.com/jupyterhub/the-littlest-jupyterhub/issues/642


meeseeksmachine:
This pull request has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/running-juptyerlab-with-jupyterhub-tljh-on-own-server/7249/7

PR #649, opened 2021-01-09
use Docker Compose for the development environment

afeld:
Combines the building of the image and the starting of the container into a single simpler command.

 - [ ] ~~Add / update documentation~~
 - [ ] ~~Add tests~~

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,11 @@
+version: "3.9"
+services:
+  hub:
+    build: ./integration-tests
+    privileged: true
+    volumes:
+      - type: bind
+        source: .
+        target: /srv/src
+    ports:
+      - "12000:80"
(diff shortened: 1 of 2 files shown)


consideRatio:
Thanks for the suggestion @afeld!

I'm thinking this adds more complexity than it benefits the maintenance efforts of this repo, and would due to that not like to merge this change. If I were more familiar with docker compose I may think it became easier, but I'm not very familiar with it to work with it at ease.

I've just spent a lot of time in the CI system and would say there is plenty of complexity in general to run our tests because of the nature of the project. It has a script to install a python package, which then itself installs apt-get and python packages etc, which then should be running successfully - it is tricky!

@GeorgianaElena @yuvipanda what do you think?

manics:
docker-compose is good when you need to orchestrate multiple containers.

There's only one container here, so could you perhaps outline the advantages and disadvantages of using docker-compose over a wrapper script that does the build and run?

jonasjancarik:
> There's only one container here, so could you perhaps outline the advantages and disadvantages of using docker-compose over a wrapper script that does the build and run?

One advantage is "declarative configuration": with docker-compose.yml, you can define the container's configuration in a file, which makes it easy to understand, version, and share the container's configuration with others. This also makes it simple to update the container's configuration and apply those changes consistently across different environments.

In our environment we tend to launch everything with `docker compose` precisely because it's easier for the next person to figure out how exactly a container was launched. When you use `docker run`, the next person may have a hard time figuring out exactly how you set the container up (they'll have to use `docker inspect` etc.)

PR #652, opened 2021-01-21
Own server install sets admin password in step 3

leouieda:
Fix the install instructions for own server setup where it said that 
the password was set in step 2 instead of 3.

 - [x] Add tests


GeorgianaElena:
Thanks for spotting and fixing this @leouieda :heart:

PR #656, opened 2021-02-04
setup.py: Update repo URL

jayvdb:


setup.py
```diff
@@ -4,7 +4,7 @@
     name='the-littlest-jupyterhub',
     version='0.1',
     description='A small JupyterHub distribution',
-    url='https://github.com/yuvipanda/the-littlest-jupyterhub',
+    url='https://github.com/jupyterhub/the-littlest-jupyterhub',
     author='Yuvi Panda',
     author_email='yuvipanda@gmail.com',
```
jtpio, line 9: Should these fields also be updated to match the ones from the `jupyterhub` project, since TLJH is also an official sub-project? 

https://github.com/jupyterhub/jupyterhub/blob/2e198396c172b6e4cdfc023466a9eb29467d0a92/setup.py#L91-L93

yuvipanda: Yeah, I agree - @jayvdb do you think you can do that? Happy to merge after.

Apologies for the late review!

jayvdb: updated author fields

yuvipanda:
Thanks, @jayvdb! The fail seems unrelated, so am going to merge this.

PR #663, opened 2021-02-15
Add docs to override systemd settings

### LINKED ISSUE #661: Provide a way to configure or remove PrivateTmp=yes in the jupyterhub systemd unit
### Proposed change

This is something that came up while trying to mount a shared volume with samba, using the following command from a python `subprocess` when the user logs in:

```bash
mount -t cifs -o user=username,password=abcde1234 //example.com/Data /mnt/username/data
```

However it looks like `systemd` namespaces the mounts, which means that the mounted volume was then not visible to the end user.

Removing `PrivateTmp=yes` from the systemd unit file fixed the issue:

https://github.com/jupyterhub/the-littlest-jupyterhub/blob/18efee665aebd6c59334685eb9929f823ad649b0/tljh/systemd-units/jupyterhub.service#L13

In this setup, the share would first be mounted under `/mnt/username/data` on the host, and then mounted into `/home/jovyan/data` in the user container (using `DockerSpawner`)

### Alternative options

For now it's possible to patch the `jupyterhub.service` systemd unit by either editing it inplace with `sed`, or resinstalling the modified unit without `PrivateTmp=yes`. 

TLJH admins that would like to mount volumes with samba.

It might also apply to other network based file systems such as NFS? (although not tested).

Not sure what the best way would be for now, but maybe this could be made optional or configurable.
Or remove `PrivateTmp=yes`.

jtpio:
Fixes #661 with documentation.

 - [ ] Add tests

image


jtpio:
cc @adriendelsalle who was helping looking into this

docs/howto/admin/systemd.rst
```diff
@@ -0,0 +1,77 @@
...
+  Environment=TLJH_INSTALL_PREFIX=/opt/tljh
+  ExecStart=/opt/tljh/hub/bin/python3 -m jupyterhub.app -f jupyterhub_config.py --upgrade-db
+
+  [Install]
+  WantedBy=multi-user.target
+
+
+However in some cases, admins might want to have better control on these settings.
+
+For example when mounting shared volumes over the network using Samba, these namespacing settings might be a bit too strict
```
GeorgianaElena, line 37: Maybe add a link to Samba here, so that people can learn more about it?

docs/howto/admin/systemd.rst
```diff
@@ -0,0 +1,77 @@
...
+and prevent users from accessing the shared volumes.
+
+
+Overriding settings with ``override.conf``
+==========================================
+
+To override the settings, it is possible to provide a custom ``/etc/systemd/system/jupyterhub.service.d/override.conf`` file
```
GeorgianaElena, line 44:
```suggestion
To override the ``jupyterhub`` settings, it is possible to provide a custom ``/etc/systemd/system/jupyterhub.service.d/override.conf`` file
```

docs/howto/admin/systemd.rst
```diff
@@ -0,0 +1,77 @@
...
+with the following content:
```
GeorgianaElena, line 45: I think we should specify that this is only an override example and not something they **must** do.
Or mention that this is relevant for the Samba scenario? What do you think?

docs/howto/admin/systemd.rst
```diff
@@ -0,0 +1,77 @@
...
+    ProtectKernelModules=no
+
+Then make sure to reload the daemon and the ``jupyterhub`` service:
+
+.. code-block:: bash
+
+    sudo systemctl daemon-reload
+    sudo systemctl restart jupyterhub
+
+Checking the status should look like the following:
```
GeorgianaElena, line 62: Maybe also provide the command to check the status? I believe it's already specified somewhere in the docs, but I think it's best to have it here also just for completeness.

GeorgianaElena (approved):
Thank you @jtpio and @adriendelsalle for figuring this out and for sharing the solution with everyone :tada:

docs/howto/admin/systemd.rst
jtpio: Sounds good!

docs/howto/admin/systemd.rst
jtpio: Good point, we can mention that it is just an example and more (or less) options could be specified in the override file.

jtpio:
Thanks @GeorgianaElena for the review!

I pushed a couple of changes to improve the documentation.

GeorgianaElena:
Thank you @jtpio :tada:

SylvainCorlay:
It seems that there are some outstanding issues with samba mounts:

 - Samba mounts disappear if server is idle for too long.
 - Samba mounts do not show up in the JupyterLab file browser despite being available from bash.

The latter may be due to the jupyter server process running with different permissions than the kernel.

PR #676, opened 2021-04-01
Add the jupyterhub-configurator service

GeorgianaElena:
Right now it only allows changing the notebook interface.

TODO
 - [ ] Add tests


tljh/configurer.py
```diff
@@ -66,6 +66,9 @@
             'concurrency': 5,
             'users': False,
             'max_age': 0
+        },
+        'configurator': {
+            'enabled': True
```
yuvipanda, line 71: let's default this to false for now?

tljh/user_creating_spawner.py
```diff
@@ -2,8 +2,9 @@
 from tljh import user
 from systemdspawner import SystemdSpawner
 from traitlets import Dict, Unicode, List
+from jupyterhub_configurator.mixins import ConfiguratorSpawnerMixin
 
-class UserCreatingSpawner(SystemdSpawner):
+class UserCreatingSpawner(ConfiguratorSpawnerMixin, SystemdSpawner):
```
yuvipanda, line 7: We should use this mixin only if configurator is enabled - otherwise, any bugs in the configurator backend will stop new user spawns!

yuvipanda:
THIS IS GREAT! I made a couple suggestions, but after that let's give this a go!

tljh/user_creating_spawner.py
GeorgianaElena: Good idea. I just added a commit that does this with [`type`](https://docs.python.org/3/library/functions.html?#type) to avoid duplicating code. I tried it within a dev environment and it worked but I'm not sure at all if this is acceptable or if there's a better way.

tljh/user_creating_spawner.py
```diff
@@ -32,3 +33,8 @@ def start(self):
                     user.ensure_user_group(system_username, group)
         return super().start()
 
+cfg = configurer.load_config()
+if cfg['services']['configurator']['enabled']:
```
yuvipanda, line 37: Very cool type-mashing! Can you leave a comment here on why we are doing this, and how it works?

tljh/schemas/tljh_configurator.py
```diff
@@ -0,0 +1,27 @@
+from jupyterhub_configurator.hooks import hookimpl
```
yuvipanda, line 1: Let's put this in just tljh/configurator_schemas.py, instead of making its own modules? In the future, we might want to use Jupyter Telemetry for something, and that will have its own schemas.

tljh/schemas/tljh_configurator.py
```diff
@@ -0,0 +1,27 @@
...
+
+@hookimpl
+def jupyterhub_configurator_fields():
```
yuvipanda, line 4: Let's put this file in https://github.com/yuvipanda/jupyterhub-configurator for now? The format of these schemas is changing pretty rapidly. We can refer to this from there!

yuvipanda:
This is great! Let's add some docs on how you can experimentally enable this, and then merge?

GeorgianaElena:
@yuvipanda, I added a section about the configurator. It's not complete, but I think it works for now.
I'm going to merge this now and see how it goes :crossed_fingers:

PR #673, opened 2021-03-24
Switch integration and upgrade tests from CircleCI to GitHub actions

GeorgianaElena:
- [x] Switch integration tests to GHA
 - [x] Remove integration tests from CircleCI
 - [x]  Switch upgrade tests to GHA
 - [x]  Remove upgrade tests from CircleCI

PR #678, opened 2021-04-02
Revert "Revert "Switch integration and upgrade tests from CircleCI to GitHub actions""

yuvipanda:
Reverts jupyterhub/the-littlest-jupyterhub#677

Brings back https://github.com/jupyterhub/the-littlest-jupyterhub/pull/673,
needs to be properly tested :(


GeorgianaElena:
@yuvipanda, this should be ready for review again.
I've get rid of the use of `GITHUB_ACTOR` and replaced that with `GITHUB_REPOSITORY` and `GITHUB_SHA` with `GITHUB_REF` (ref [GHA default env vars](https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables)). This should make the tests run ok both on `pull_request` events as well as `push` events.

yuvipanda:
ok, let's try it!

Issue #685, opened 2021-05-04
`jupyterhub-nativeauthenticator v0.0.7` breaks the sign up page.


ikhoury:
### Bug description
Sign up page returns `500 internal server error`. Users are unable to sign up to the service.

#### Expected behaviour
Sign up page should load correctly.

#### Actual behaviour
Sign up page returns `500 internal server error`.

### How to reproduce

1. Go to your JupyterHub login page.
2. Click on 'Sign Up'.

### Your personal set up
Single server `the-littlest-jupyterhub` installation.

 - OS: Ubuntu 18.04
 - Version(s): JupyterHub `1.2.2`, Python `3.6.9`

- <details><summary>Full environment</summary>

```
aiohttp==3.7.4.post0
alembic==1.6.0
async-generator==1.10
async-timeout==3.0.1
attrs==20.3.0
backoff==1.10.0
bcrypt==3.2.0
certifi==2020.12.5
certipy==0.1.3
cffi==1.14.5
chardet==3.0.4
cryptography==3.4.7
decorator==5.0.7
deepmerge==0.3.0
entrypoints==0.3
escapism==1.0.1
etcd3==0.12.0
greenlet==1.0.0
grpcio==1.37.1
idna==2.10
idna-ssl==1.1.0
importlib-metadata==4.0.1
ipython-genutils==0.2.0
Jinja2==2.11.3
jsonschema==3.2.0
jupyter-telemetry==0.1.0
jupyterhub==1.2.2
jupyterhub-configurator==1.0
jupyterhub-dummyauthenticator==0.3.1
jupyterhub-firstuseauthenticator==0.14.1
jupyterhub-idle-culler==1.0
jupyterhub-ldapauthenticator==1.3.0
jupyterhub-nativeauthenticator==0.0.7
jupyterhub-systemdspawner==0.15.0
jupyterhub-tmpauthenticator==0.6
jupyterhub-traefik-proxy==0.2.0
ldap3==2.9
Mako==1.1.4
MarkupSafe==1.1.1
multidict==5.1.0
oauthenticator==0.10.0
oauthlib==3.1.0
onetimepass==1.0.1
pamela==1.0.0
passlib==1.7.4
pkg-resources==0.0.0
pluggy==1.0.0.dev0
prometheus-client==0.10.1
protobuf==3.15.8
pyasn1==0.4.8
pycparser==2.20
pycurl==7.43.0.6
pyOpenSSL==20.0.1
[...]

PR #688, opened 2021-05-10
Bump hub and notebook versions

GeorgianaElena:
Fixes https://github.com/jupyterhub/the-littlest-jupyterhub/issues/685


jihobak:
@GeorgianaElena Thanks for the updated. When can we use official updated version ? :)

PR #704, opened 2021-08-30
Bump nbgitpuller version

yuvipanda:
Brings in a security fix.

See https://github.com/jupyterhub/nbgitpuller/security/advisories/GHSA-mq5p-2mcr-m52j

 - [x] Add tests


consideRatio:
@yuvipanda, do you think this will work fine without further changes to be documented by updating to nbgitpuller 1.0.2? If so, this LGTM and can be merged.

PR #710, opened 2021-10-16
Revision of our GitHub Workflows and README.rst to README.md

consideRatio:
### PR STATUS

Ready for review / merge in my mind. I gave up on getting the unit test to work against 21.10 and added a note about it instead pointing to #714 that remains open.

### Changes made summarized

- I found the workflow files hard to read due to a lack of vertical spacing. I noted some trailing `\` on script lines without a line following them. And I noted that pytest was run without --color=yes as needed in a GitHub workflow.
- I also put a limit on the maximum failures in pytest to be 2, because its often hard to read logs if there are multiple failures, yet it can be relevant to see if there was more than one failure - so having two is a reasonable compromise we've used in other projects.
- The changes in the readme are:
  1. rST -> Markdown
  2. Circle CI badge removed, typical jh org repo badges added
- I noted that the integration-test and upgrade-test workflow were the same besides a single flag. I understand the reason for having them in separate workflows has been that one could not declare a matrix that only sometimes run one of the variations. I found a workaround to this problem by adding a small job calculating the matrix dynamically ahead of the real test job starts. By doing this, I could remove the duplicated test logic in the upgrade-test workflow.
- We test against ubuntu 18.04, 20.04, Python 3.6, Python 3.9, and with --upgrade.
  - Yuvi figured out what took me ages to figure out in #674, that /lib/systemd/systemd is needed as an entrypoint in 20.04.
  - I consider this PR to be closing #674 as it mostly includes it
  - I've failed to run unit tests against ubuntu 21.10, the reason for failing is described in #714.
- We bump traefik-proxy to 0.3.0 to include a bugfix and optional install of etcd3 python package that caused installation trouble
- We bump pip to 21.3.* from 0.20.*, I think this was needed for the integration tests to succeed.

### Post merge action point

- [x] Update the merge restrictions to reflect the new job names if needed in https://github.com/jupyterhub/the-littlest-jupyterhub/settings/branch_protection_rules/2098820.
  image
  
### Related

- Almost closes the stale #521 PR that updates CircleCI config, but perhaps not fully as we don't test against 21.10 for example.
- Closes #381 by testing against 18.04 and 20.04

### DIFF
--- a/.github/workflows/integration-test.yaml
+++ b/.github/workflows/integration-test.yaml
@@ -102,23 +102,13 @@ jobs:
         with:
           python-version: ${{ matrix.python_version }}
 
-      - name: Set BOOTSTRAP_PIP_SPEC value
-        run: |
-          echo "BOOTSTRAP_PIP_SPEC=git+https://github.com/$GITHUB_REPOSITORY.git@$GITHUB_REF" >> $GITHUB_ENV
-          echo $BOOTSTRAP_PIP_SPEC
-
-      - name: Build systemd image
-        run: |
-          .github/integration-test.py build-image \
-              --build-arg "ubuntu_version=${{ matrix.ubuntu_version }}"
-
       - name: Install pytest
         run: python3 -m pip install pytest
 
       # We abort pytest after two failures as a compromise between wanting to
       # avoid a flood of logs while still understanding if multiple tests would
       # fail.
-      - name: Run bootstrap tests
+      - name: Run bootstrap tests (Runs in/Builds ubuntu:${{ matrix.ubuntu_version }} derived image)
         run: |
           pytest --verbose --maxfail=2 --color=yes --durations=10 --capture=no \
               integration-tests/test_bootstrap.py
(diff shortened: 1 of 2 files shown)


consideRatio:
@yuvipanda :sob: there is another test failure, probably related to traefik-proxy version bump.

It seems intermittent as well, I'm not sure what's going on. I'll take a break from this PR for a while, I'm so tired of seeing these tests fail over and over :D Here is a changelog for traefik-proxy btw if you feel like debugging it: https://github.com/jupyterhub/traefik-proxy/blob/master/docs/source/changelog.md

---

@yuvipanda this PR is flooded with commits and it made a comment minimized for me - did you see this response?

https://github.com/jupyterhub/the-littlest-jupyterhub/pull/710#discussion_r731216789

I was about to summarize why we pin pip in a comment and describe that we should update every three/six months, but I realize we haven't really motivated that in our discussion. If we should cut an update directly when new releases comes, are we pinning motivated by something else than this?
- to make sure someone running bootstrap script for the first time gets a tested installation
- to make sure someone running bootstrap script twice don't get an untested pip version (released every three months)

consideRatio:
Oooooh noooo, the test failure is even intermittent it seems!!! Noooooooooooooooooooooooooooooooo!!

yuvipanda:
@manics the test failures here seem to be when downloading traefik rather than conda. I cleaned up the conda downloader code you mentioned in https://github.com/jupyterhub/the-littlest-jupyterhub/pull/717.

We have our own code in TLJH to download the traefik binary - perhaps we should move it to use the jupyter-traefik-proxy downlader code instead?

consideRatio:
Yes! @GeorgianaElena has opened a PR about using ubuntu 20.04 i think, marked as WIP (im on mobile).

PR #342, opened 2019-05-20
Add instructions to deploy on Azure

trallard:
- [x] Add / update documentation
 - [ ] Add tests

 

This PR adds the instructions to set a suitable VM for the littles JupyterHub on Azure.

PR #713, opened 2021-10-17
Remove template in root folder - a mistakenly committed file

consideRatio:
I believe the file named "template" in the root directory of this project was added by mistake in https://github.com/jupyterhub/the-littlest-jupyterhub/pull/342.

@trallard do you remember if this file has a purpose, or if it could have been a file added by mistake?


trallard:
Honestly I cannot remember - so it might have been added by mistake so it should be good to remove

PR #748, opened 2021-10-28
Add .pre-commit-config

consideRatio:
I've applied all but the black autoformatting changes for this pre-commit-config in #747. The config contains typical pre-commit hooks as found in jupyterhub, kubespawner, and other repos.


yuvipanda:
Thanks!

PR #754, opened 2021-11-01
pre-commit: remove requirements-txt-fixer

consideRatio:
I mistakenly added requirements-txt-fixer which has turned out to cause problems, which is in this PR documented in the relevant requirements.txt file to help figuring out why that is in more detail at some later time.

### DIFF
--- a/tljh/requirements-base.txt
+++ b/tljh/requirements-base.txt
@@ -1,9 +1,10 @@
 # When tljh.installer runs, the users' environment as typically found in
 # /opt/tljh/user, is setup with these packages.
 #
-# FIXME: a frozen version of this file should be used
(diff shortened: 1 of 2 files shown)


tljh/requirements-base.txt
```diff
@@ -1,9 +1,10 @@
 # When tljh.installer runs, the users' environment as typically found in
 # /opt/tljh/user, is setup with these packages.
 #
-# FIXME: a frozen version of this file should be used
-# pinning only direct dependencies is a recipe for broken environments!
-
```
consideRatio, line 6: Oh, I guess I also removed this comment in this PR =/

This was there when all packages were pinned strict.

PR #755, opened 2021-11-01
pre-commit: apply black formatting (and prettier on one yaml file)

consideRatio:
I've applied black formatting now. I ran it once first without string normalization, applied some manual tweaks, and then ran it with string normalization.

String normalization isn't systematically used with black in the jupyterhub org. I personally favor it though and figure it could make sense to do while also applying black in the first place - something that is reasonable to do only when the amount of PRs are low.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -87,7 +87,6 @@ commands:
           command: |
             py.test integration-tests/test_bootstrap.py -s
 
-
 jobs:
   integration-test:
     docker:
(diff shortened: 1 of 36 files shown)


consideRatio:
@manics @yuvipanda @choldgraf what do you think?

The PR list is down from 20 to 1 single stale PR with very little changes, can we go for a full `black` CLI transform at this point?

Waiting for this to get merged is somewhat of a blocker to start doing more work as well.

image

choldgraf (approved):
I'm +1 on merging this - the remaining PR isn't complicated enough that it would be majorly disrupted by this. I think it'd be OK to merge so that we don't start to create conflicts when new PRs come up

manics (approved):
LGTM!

consideRatio:
Thanks @manics @choldgraf! I'll go for a merge

Issue #770, opened 2021-12-16
outdated docs on Get System Log for AWS EC2


raybellwaves:
Section 18 here (https://tljh.jupyter.org/en/latest/install/amazon.html) says: Right-click Instance Settings > Get System Log.

The buttons have changed so it's now Monitor and troubleshoot -> Get system log

![Screen Shot 2021-12-16 at 2 45 25 PM](https://user-images.githubusercontent.com/17162724/146440021-eb25d897-d03b-4705-bf74-80772003b6e6.png)

PR #772, opened 2021-12-17
DOC: update info on AWS get system log

raybellwaves:
Closes https://github.com/jupyterhub/the-littlest-jupyterhub/issues/770

I opted not to update the screen shot but I can do if required.


GeorgianaElena (approved):
Thanks @raybellwaves! :sparkles:

Issue #773, opened 2021-12-17
launch on /lab by default


raybellwaves:
### Proposed change
When launching tljh for the first time in launches using /lab instead of /tree (classic) (

### Alternative options
You can do it by hand after launch by running
`sudo tljh-config set user_environment.default_app jupyterlab`

I also don't know how to pass this on at the build stage. e.g. to pass as an arg to
```
#!/bin/bash
curl -L https://tljh.jupyter.org/bootstrap.py \
  | sudo python3 - \
    --admin <user-name>
```

### Who would use this feature?
Hopefully everyone as lab is more convient for dropping into a terminal for example to run `sudo tljh-config` commands

### (Optional): Suggest a solution

Change https://github.com/jupyterhub/the-littlest-jupyterhub/blob/main/tljh/configurer.py#L52 to jupyterlab

May be something here as well https://jupyterlab.readthedocs.io/en/stable/user/jupyterhub.html#use-jupyterlab-by-default

raybellwaves:
I learnt it's easy to do currently via a plugin e.g. https://github.com/yuvipanda/tljh-pangeo/blob/master/tljh_pangeo.py#L14-L22

PR #774, opened 2021-12-17
launch on lab by default

raybellwaves:
- [ ] Add tests

 

Related to https://github.com/jupyterhub/the-littlest-jupyterhub/issues/773.

I'm not sure the mappings between `default_app` and `default_url`. Am I right that there is also `jupyter-labhub` e.g. https://github.com/manics/jupyterhub-rtc-example/blob/main/jupyterhub_config.py#L33

### DIFF
--- a/tljh/configurer.py
+++ b/tljh/configurer.py
@@ -49,7 +49,7 @@ default = {
         "password": "",
     },
     "user_environment": {
-        "default_app": "classic",
+        "default_app": "jupyterlab",
     },
     "services": {
         "cull": {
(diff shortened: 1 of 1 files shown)


tljh/configurer.py
```diff
@@ -228,6 +228,8 @@ def update_user_environment(c, config):
     # Set default application users are launched into
     if user_env["default_app"] == "jupyterlab":
         c.Spawner.default_url = "/lab"
+    elif user_env["default_app"] == "classic":
+        c.Spawner.default_url = "/tree"
```
raybellwaves, line 232: I think this causes this test to fail (https://github.com/jupyterhub/the-littlest-jupyterhub/blob/main/tests/test_configurer.py#L55) by giving it a default _url.

raybellwaves:
Closing in favor of https://github.com/jupyterhub/the-littlest-jupyterhub/pull/775

Issue #724, opened 2021-10-22
Handling versions and releases


consideRatio:
I've grown to think that cutting releases is a way to reduce maintenance workload, rather than increase it. By having releases, one can understand and pinpoint issues a lot easier, and communicate about changes to the end users that can figure out their own issues a lot easier.

I'm thinking it would be good if TLJH started handling releases. Do I have support for this idea if I or someone else would start to work towards that?

### The Bootstrap script

We currently have a bootstrap/bootstrap.py script that downloads the default branch from github. I suggest that script is remade to download a given version, which would be the latest if it is unspecified.

We have documented access to this script like...

```
curl -L https://tljh.jupyter.org/bootstrap.py
```

So in practice, users can perhaps always be assumed to use the latest version of the script, but we can still add a section in the TLJH changelog about changes to this script over time.

### The TLJH Python package

I suggest we version control this as any other Python package.

### Related

- About defining an upgrade strategy, but then we need version to relate to: https://github.com/jupyterhub/the-littlest-jupyterhub/issues/81

manics:
A consequence of _not_ having releases is
```
root@ubuntu1804:~# /opt/tljh/hub/bin/pip list | grep the-littlest-jupyterhub
the-littlest-jupyterhub          0.1
```
will always be 0.1.

This makes debugging upgrades quite difficult, e.g. https://github.com/jupyterhub/the-littlest-jupyterhub/issues/762

jtpio:
Thanks @consideRatio for raising this :+1: 

This sounds particularly useful especially now that JupyterHub 2.0 is out.

TLJH is still pinned on 1.* for now:

https://github.com/jupyterhub/the-littlest-jupyterhub/blob/ad38fad96afe8096e722279da095a0633b5d2c6b/tljh/installer.py#L124

According to the changelog we can expect a couple of breaking changes: https://jupyterhub.readthedocs.io/en/stable/changelog.html?#id1

Which would affect downstream users and plugins if they just run the installer to upgrade.

Maybe the Jupyter Releaser would be able to help here? https://github.com/jupyter-server/jupyter_releaser

manics:
Two ideas:

## Version the installer URL
Include a version/branch component in the URL to the bootstrap script, e.g.

`https://tljh.jupyter.org/v2.1.0/bootstrap.py`
`https://tljh.jupyter.org/v2/bootstrap.py`

[...]

PR #775, opened 2021-12-17
Launch into `/lab` by default by changing TLJH config's default value

raybellwaves:
This PR changes the default value for the TLJH config `user_environment.default_app` to `jupyterlab`, which makes `c.Spawner.default_url` be set to `/lab` by default in TLJH.

To revert to the behavior before this, a TLJH admin can do:

```
sudo tljh-config set user_environment.default_app classic
```

---

- [x] Add tests

Closes https://github.com/jupyterhub/the-littlest-jupyterhub/issues/773

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -74,6 +74,32 @@ myst_enable_extensions = [
     "fieldlist",
 ]
 
+
+# -- Options for intersphinx extension ---------------------------------------
+# ref: https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html#configuration
+#
+# The extension makes us able to link like to other projects like below.
+#
+#     rST  - :external:py:class:`jupyterhub.spawner.Spawner`
+#     MyST - {external:py:class}`jupyterhub.spawner.Spawner`
+#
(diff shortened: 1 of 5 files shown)


GeorgianaElena (commented):
Thanks for opening this @raybellwaves!

I don't feel comfortable enough to merge this, because I believe this needs some discussion before. For example, there is a [conversation about this in `z2jh`](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/776) which I believe concluded that migrating to JupyterHub 2.0 which defaults to JupyterLab is the way to go.

Maybe we can do the same for TLJH and focus on migrating to JupyterHub 2.0, and then figure what needs to change in order to accommodate the hub's default interface.

manics:
If we resolve https://github.com/jupyterhub/the-littlest-jupyterhub/issues/724 this should be easier to deal with

consideRatio:
I rebased this PR and pushed a commit updating the docs about this change, and adding a breaking change changelog entry about it.

consideRatio (approved):

minrk (approved):

PR #776, opened 2021-12-17
add info line saying TLJH setup complete

raybellwaves:
When looking at the logs i'm not sure when the setup has finished as the last line is `logger.info("Running TLJH installer...")`. This PR adds the line `logger.info("TLJH setup complete")` so myself and others know we can go have fun

### DIFF
--- a/bootstrap/bootstrap.py
+++ b/bootstrap/bootstrap.py
@@ -389,7 +389,8 @@ def main():
     # Run TLJH installer
     logger.info("Running TLJH installer...")
     os.execv(python_bin, [python_bin, "-m", "tljh.installer"] + tljh_installer_flags)
-
+    logger.info("TLJH setup complete")
+    
 
 if __name__ == "__main__":
     main()


raybellwaves:
I also see

```
2021-12-17 04:31:59,773 Waiting for JupyterHub to come up (1/20 tries)
2021-12-17 04:31:59,788 Done!
```
in the logs so this may not be needed

consideRatio:
The added code is unreachable because of the call to `os.execv`:

> These functions all execute a new program, replacing the current process; they do not return. On Unix, the new executable is loaded into the current process, and will have the same process id as the caller. Errors will be reported as [OSError](https://docs.python.org/3/library/exceptions.html#OSError) exceptions.

From https://docs.python.org/3/library/os.html

---

> I also see in `opt/tljh/installer.log`
> 
> ```
> 2021-12-17 04:31:59,773 Waiting for JupyterHub to come up (1/20 tries)
> 2021-12-17 04:31:59,788 Done!
> ```
> 
> in the logs so this may not be needed

Excellent, closing!

PR #777, opened 2021-12-17
bump jupyterhub to 2.*

raybellwaves:
- [ ] Add tests

 

Firstly, does it still work?
Secondly, any reason why not?


manics:
I think we'll need to resolve https://github.com/jupyterhub/the-littlest-jupyterhub/issues/724 since this is a breaking change

jtpio:
This will fix https://github.com/jupyterhub/the-littlest-jupyterhub/issues/818

consideRatio:
- Closed by #842

PR #551, opened 2020-04-20
Add a section about known TLJH plugins to the documentation

jtpio:
Fixes #550.

This change adds a section about known TLJH plugins to the documentation:

image

 - [x] Add / update documentation
 - [x] ~Add tests~

PR #781, opened 2021-12-19
DOC: move link to contributing/plugin higher

raybellwaves:
I'll confess I didn't find this section until I came across https://github.com/jupyterhub/the-littlest-jupyterhub/pull/551

This PR moves the link higher up in the section to get more visibility. It also gives more context about what is on that page.

I also made a note that `git+https://github.com/yuvipanda/tljh-pangeo@v0.1` doesn't exist as I may not be the only person to copy and paste and get an error.


GeorgianaElena (approved):
Good point @raybellwaves! Thank you!

PR #784, opened 2021-12-20
Specify user-environment Python version

oisinBates:
- [ ] Add tests

I  am suggesting this feature as I think that it's useful to enable the ability to specify a specific Python version for the User Environment, and this is a specific pain point that I encountered.


consideRatio:
I'm very happy to see a feature like this be added. I'd like to see it scoped to only be functional during initial install and setup of the user environment though, as its a bit too intrusive to upgrade an existing Python version in the user's environment.

PR #785, opened 2021-12-21
DOC: update sudo tljh-config --help demo

raybellwaves:
The current example (https://tljh.jupyter.org/en/latest/topic/tljh-config.html#advanced-config-yaml) doesn't show `unset`


GeorgianaElena (approved):
Great catch @raybellwaves! Thank you!

PR #789, opened 2021-12-29
extra logger.info

raybellwaves:
Adding this as I need to check i'm passing the environment variable correctly


yuvipanda:
Thanks @raybellwaves!

PR #800, opened 2022-02-20
Add support for debian >=10 to bootstrap.py

jochym:
- [ ] Add tests

### DIFF
--- a/bootstrap/bootstrap.py
+++ b/bootstrap/bootstrap.py
@@ -130,7 +130,6 @@ progress_page_html = """
 
 logger = logging.getLogger(__name__)
 
-
 # This function is needed both by the process starting this script, and by the
 # TLJH installer that this script execs in the end. Make sure its replica at
 # tljh/utils.py stays in sync with this version!
@@ -199,11 +198,14 @@ def ensure_host_system_can_install_tljh():
     # Require Ubuntu 18.04+
     distro = get_os_release_variable("ID")
     version = float(get_os_release_variable("VERSION_ID"))
-    if distro != "ubuntu":
-        print("The Littlest JupyterHub currently supports Ubuntu Linux only")
+    if distro in ["ubuntu", "debian"]:
+        print("The Littlest JupyterHub currently supports Ubuntu or Debian Linux only")
         sys.exit(1)
-    elif float(version) < 18.04:
+    elif distro == "ubuntu" and float(version) < 18.04:
         print("The Littlest JupyterHub requires Ubuntu 18.04 or higher")
+        sys.exit(1)
+    elif distro == "debian" and float(version) < 10:
+        print("The Littlest JupyterHub requires Debian 10 or higher")
         sys.exit(1)
 
     # Require Python 3.6+
@@ -224,6 +226,7 @@ def ensure_host_system_can_install_tljh():
                 "For local development, see http://tljh.jupyter.org/en/latest/contributing/dev-setup.html"
             )
         sys.exit(1)
+    return distro, version 
 
 
 class ProgressPageRequestHandler(SimpleHTTPRequestHandler):
@@ -259,7 +262,7 @@ def main():
     start a local webserver temporarily and report its installation progress via
     a web site served locally on port 80.
     """
-    ensure_host_system_can_install_tljh()
+    distro, version = ensure_host_system_can_install_tljh()
 
     # Various related constants
     install_prefix = os.environ.get("TLJH_INSTALL_PREFIX", "/opt/tljh")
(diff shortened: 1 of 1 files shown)


manics:
Personally I have no objections to adding other distributions providing they're tested and easy to maintain- in this case I'd probably go for officially supporting and testing Debian 11 and not 10. However I'd like input from the other maintainers, especially @yuvipanda @consideRatio, first.

I think it'd be good to sort out https://github.com/jupyterhub/the-littlest-jupyterhub/issues/724 before making any significant changes to TLJH, though we could perhaps merge this PR with tests but make clear it's not supported for now?

consideRatio:
The similarities between Debian / Ubuntu makes for a good case of supporting both if there is demand and adds value. I think that #724 is a critical issue to resolve that would make me more comfortable considering features overall.

While this now installs okay and that is great, it adds complexity. Suddenly we have documentation that may be outdated for debian by assumptions of ubuntu and get questions about that - maybe not - I'm not sure. We have CI pipelines that take a bit longer to run because we test more variations I assume. 

But since there is no real change, I think I'm inclined to agree that supporting debian is acceptable if:
1. The change in this code base is almost only a matter of relaxing constraints forbidding anything but ubuntu to also accept debian.
2. If we print a warning that we haven't developed this to be installed in debian, so docs may be ubuntu specific.
3. If we clarify with an inline comments that `sudo` now is installed explicitly because its not arriving by default with debian, but does with ubuntu.
4. Two additional JupyterHub team members agree besides me and there isn't any further pushback from another member.

@GeorgianaElena @minrk @yuvipanda what do you think?

bootstrap/bootstrap.py
```diff
@@ -345,7 +348,8 @@ def serve_forever(server):
             ["apt-get", "install", "--yes", "software-properties-common"],
             env=apt_get_adjusted_env,
         )
-        run_subprocess(["add-apt-repository", "universe", "--yes"])
+        if distro == "ubuntu":
+            run_subprocess(["add-apt-repository", "universe", "--yes"])
```
consideRatio, line 352: Can you clarify and inline a comment about why this is conditional to ubuntu?

bootstrap/bootstrap.py
```diff
@@ -356,6 +360,7 @@ def serve_forever(server):
                 "python3-venv",
                 "python3-pip",
                 "git",
+                "sudo",
```
consideRatio, line 363: I assume `sudo` is installed in ubuntu by default, but needs to be installed in debian - is that correct? If so, let's add an inline comment about it!

bootstrap/bootstrap.py
jochym: Universe section is specific to Ubuntu, so it cannot be added in Debian. And it is not needed there since all required packages are in main. Probably, the corresponding section in Debian would be "contrib".

bootstrap/bootstrap.py
jochym: Sudo is. indeed, not installed in Debian by default. But adding it here is harmless if it is installed and fixes it for Debian.

manics: I think this is OK, even if it's pre-installed in Ubuntu this acts as documentation that sudo is required. It's also useful if someone uses a custom Ubuntu cloud image with minimal dependencies.

On the other hand https://tljh.jupyter.org/en/latest/install/custom-server.html#step-1-installing-the-littlest-jupyterhub mentions `sudo apt install python3 python3-dev git curl` is a pre-requisite so we should decide what is a prerequisite, and what should be managed by `bootstrap.py`

jochym: You are right. But since I needed to modify the script and I never run scripts from the pipe to sudo I downloaded the script with wget and run it with python. If the sudo was not installed previously, the installation would fail. 
Note that, there is no harm in requiring installation of already installed component.

manics: OK, presumably you're running `bootstrap.py` as `root` then? That's why I asked for your workflow :smile: 

It looks like the sudo package is only required for some config files as far as the installer is concerned:
https://github.com/jupyterhub/the-littlest-jupyterhub/blob/0572976d25c18bc88d936f8bd8627bbfed069b53/tljh/installer.py#L147-L153

jochym: Yes, I was running it from root. Sudo is required in ubuntu install and probably (?) during some admin tasks. Debian is not using it for admin, but it is fully supported there.

jochym:
@yuvipanda I added the requested comments, but three integration tests fail for unconnected reasons on newer ubuntus. I do not get why. My fork is synced with upstream recently and have no unconnected changes. Is this some problem with upstream?
Could you take a look?

jochym:
@yuvipanda, @consideRatio : after syncing with the recent development, all the test run fine. Since I think I have included all your comments into the fork. Maybe the PR can be merged now?

consideRatio (approved):
The changes required are minimal and the similarities between debian/ubuntu are minimal overall I think. Due to that, I think this is good to go!

PR #801, opened 2022-02-20
github workflow: echo $BOOTSTRAP_PIP_SPEC

manics:
`$GITHUB_ENV` has no effect until the next step, so without this `echo $BOOTSTRAP_PIP_SPEC` is always empty


GeorgianaElena:
Thanks @manics!

PR #803, opened 2022-03-14
Add custom Sphinx ext - substitute common values in docs

### LINKED ISSUE #720: Update documentation regarding 18.04 + 20.04 supported/recommended
It seems like we support running TLJH on both 18.04 and 20.04 currently, and I think its reasonable to support the two latest LTS releases. Should we aim for that?

I figure there is docs to update with regards to this, and this issue represents that.

iameskild:
- [ ] Add tests 
  - Not added, unsure how to test Sphinx extensions.

This is an attempt to reduce the number of places certain values in the docs need to be update, i.e. supported Ubuntu versions. Currently I have only updates the references in a few places.

This Sphinx extension mechanism can be used for other values that need updating on a regular basis.

Closes #720.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,4 +1,5 @@
 import os
+import sys
 
 source_suffix = [".rst"]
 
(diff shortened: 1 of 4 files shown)


yuvipanda:
Sorry for the really late response, @iameskild! I think leaving them as is is clearer and better, rather than trying to centralize it - makes it slightly harder to read. I'm going to close this one out. 

Again, apologies for the super delayed response here.

PR #816, opened 2022-05-05
clarify direction of information in idle-culler

minrk:
JupyterHub does not poll servers, servers post activity


GeorgianaElena (approved):
Thanks @minrk!

Issue #714, opened 2021-10-17
Make unit tests against 21.10 function


consideRatio:
### Updated issue

We have integration tests against 21.10 now and they work well, but we don't have functional unit tests. Let's close this issue if we get those up and running! The current unit test failure observed is documented in the comment below: https://github.com/jupyterhub/the-littlest-jupyterhub/issues/714#issuecomment-945154101

### Original issue

I'd like to see a test against the latest Ubuntu release, such as 21.10 in this case.

I suggest this is done by first merging #710 of the three PRs adding tests for various ubuntu versions, and then opening a new PR to test against 21.10.

- #710
- #521
- #381

consideRatio:
I've tracked down the test failure listed above as a one that is fixable by releasing jupyterhub-traefik-proxy that already contains a fix. See https://github.com/jupyterhub/traefik-proxy/issues/136.

consideRatio:
Okay I give up, I can't get past https://github.com/jupyterhub/the-littlest-jupyterhub/issues/714#issuecomment-945154101. I've disabled the unit test against 21.10 in #710 and made a note about this issue and the comment above where it is described.

PR #817, opened 2022-05-09
ci: run int. and unit tests on 22.04 LTS + py3.10

MridulS:
I tried to install tljh 22.04 on a clean digital ocean machine with Ubuntu 22.04 LTS and it worked without any issues!

Just trying to run the tests to see if there are any failures with the 22.04 + py3.10 combo.

Maybe we can close https://github.com/jupyterhub/the-littlest-jupyterhub/issues/714 if this passes?


wjcapehart:
> I tried to install tljh 22.04 on a clean digital ocean machine with Ubuntu 22.04 LTS and it worked without any issues!
> 
> 
> 
> Just trying to run the tests to see if there are any failures with the 22.04 + py3.10 combo.
> 
> 
> 
> Maybe we can close https://github.com/jupyterhub/the-littlest-jupyterhub/issues/714 if this passes?

My machine was an upgrade from 20.04 LTS to 22.04 LTS, not a clean fresh build.

Bill

MridulS:
The test failures are due to `ubuntu-22.04` not being "released" properly I think, docker seems to be missing from the environment https://github.com/actions/virtual-environments/issues/5490

consideRatio:
@MridulS thanks! I'd like to drop testing against 21.10 as part of testing against 22.04. I wanted to add 21.10 as a step towards getting it to function in 22.04 when that was released, rather than test against that and other half year releases.

consideRatio:
I cleaned up the git history in this PR and pushed one commit besides the first one you had. I figure if we are to sort out the test failures (#830). Let's start from this situation when we test against 18.04, 20.04, 22.04 and go from there.

Issue #828, opened 2022-08-30
Define conda channel in tljh_extra_user_conda_packages hook


JudoWill:
### Proposed change
Enable the definition of custom conda channels in the `tljh_extra_user_conda_packages` plugin hook. I often need to install conda packages from the `bioconda`, `fastai`, or local channels.

There is even a note on line 111 of `tljh/conda.py` about making this customizable. I'm not sure of the most elegant way to adjust the code or configuration option to make the feature possible.

I could imagine new hook that allows the plugin to define the conda channels to use during install.

### Alternative options
There could be a hacky way to use the  `tljh_post_install` hook to run a second conda install on the user environment, but this seems less ideal.

I also tried returning a list of conda packages from `tljh_extra_user_conda_packages` that prefixed the channels like so:

```python
def tljh_extra_user_conda_packages():
    return ['-c', 'bioconda', 'snakemake']
```
which could add the channel to the conda command.
However, line: 372 of `tlhj/installer.py` is

```python
conda_packages = list(set(itertools.chain(*hook.tljh_extra_user_conda_packages())))
```

The `set` operation destroys the order. One could remove the `set` or replace it with a `collections.OrderedDict` to ensure the order is preserved while still de-duplicating. That would allow the proper command to be created

```bash
{conda} install -c conda-forge --json --prefix {abspath} -c bioconda snakemake
```

### Who would use this feature?
Bioinformatics researchers who provision TLHJ as a teaching tool or sharable workstation.

This is similar to #779 which suggests allowing an `env.yaml` file for user requirements. It suggests that the same functionality is possible through plugins .... except if you need to define a new channel.

PR #829, opened 2022-09-02
Add the ability to define conda channels in plugins

JudoWill:
- [x] Add tests

This PR addresses issue #828. It adds a `tljh_extra_user_conda_channels` hook to allow plugin developers to change the conda channels during user environment install. Adding an extra hook seems to be the most elegant solution as it allows install logic to be encapsulated in the plugin as opposed to requiring the end-user to add it during the install command.

I've added unit-tests, integration tests, and a note to the documentation.  If there are any stylistic changes you'd like, I'd be happy to adjust them.

### DIFF
--- a/tljh/installer.py
+++ b/tljh/installer.py
@@ -370,13 +370,16 @@ def run_plugin_actions(plugin_manager):
 
     # Install conda packages
     conda_packages = list(set(itertools.chain(*hook.tljh_extra_user_conda_packages())))
+    conda_channels = list(itertools.chain(*hook.tljh_extra_user_conda_channels()))
+    if len(conda_channels) == 0:
+        conda_channels = ('conda-forge',)
     if conda_packages:
         logger.info(
             "Installing {} user conda packages collected from plugins: {}".format(
                 len(conda_packages), " ".join(conda_packages)
             )
         )
-        conda.ensure_conda_packages(USER_ENV_PREFIX, conda_packages)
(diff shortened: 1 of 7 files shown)


JudoWill:
Hmm, the integration tests seem to be failing for reasons outside of my changes but the unit-tests I added pass fine.

Any thoughts?

yuvipanda:
Apologies for the really late response, @JudoWill. I think this is a good and useful hook to add. I don't seem to have the rights to update this PR, but I've updated this to work with the latest master in https://github.com/jupyterhub/the-littlest-jupyterhub/pull/942 (while preserving credit for your commits). I will work on merging that if the tests pass.

tljh/installer.py
```diff
@@ -370,13 +370,16 @@ def run_plugin_actions(plugin_manager):
...
+    if len(conda_channels) == 0:
+        conda_channels = ('conda-forge',)
     if conda_packages:
         logger.info(
             "Installing {} user conda packages collected from plugins: {}".format(
                 len(conda_packages), " ".join(conda_packages)
             )
         )
-        conda.ensure_conda_packages(USER_ENV_PREFIX, conda_packages)
+        conda.ensure_conda_packages(USER_ENV_PREFIX, conda_packages, conda_channels=conda_channels)
```
yuvipanda, line 382: This is the cause of the failing tests - it's passed as `conda_channels` here but is defined as `channels`

Issue #832, opened 2022-09-19
normalize function creates illegal username when hooked into active directory


kenji-monolith:
https://github.com/jupyterhub/the-littlest-jupyterhub/blob/e7728bc1b23884cf50d452d7d165c0f56421643b/tljh/normalize.py#L18

I'm getting a username in the form of 'firstname lastname' (with a space included) back from Azure AD, which results in an error when user.py tries to create a linux user in ensure_user.  I'm not sure if this is an issue with my AD setup, but in any case it seems like the normalizer should not return usernames with spaces.

David-:
My Ubuntu Server is added to the Active Directory.
I've managed to log in on the system with the pre-2000 username (DOM\DD) but not with the post-2000 username (d.david@domaine.example succeed with pam_sss but wait for 90 seconds and then failed, I can't still find why).
So, I tried it on my JupyterHub, and it begin to works (if I disable apparmor) but create the system user `jupyter-dom\dd` and the `\` is bad supported!
```
Apr 19 13:27:28 jupyter.domaine.example systemd[1]: jupyter-dom\dd.service: Failed to load environment files: No such file or directory
Apr 19 13:27:28 jupyter.domaine.example systemd[1]: jupyter-dom\dd.service: Failed to run 'start' task: No such file or directory
Apr 19 13:27:28 jupyter.domaine.example systemd[1]: jupyter-dom\dd.service: Failed with result 'resources'.
Apr 19 13:27:28 jupyter.domaine.example systemd[1]: Failed to start /opt/tljh/user/bin/jupyterhub-singleuser.
```
And when I try to delete this user in the hub admin panel, the API give a 404 because de \ is converted in a /:
`python3[738]: [W 2024-04-19 13:45:58.610 JupyterHub log:192] 404 DELETE /hub/api/users/dom/dd?_xsrf=[secret]`

I suppose the \ character should be replaced, too.

PR #833, opened 2022-09-20
replace spaces by dashes when normalizing usernames

kamatsuoka:


kamatsuoka:
This is for https://github.com/jupyterhub/the-littlest-jupyterhub/issues/832

tljh/normalize.py
```diff
@@ -8,14 +8,15 @@ def generate_system_username(username):
...
+    If username < 26 char, we just return it
+    (with any spaces replaced by dashes).
     Else, we hash the username, truncate username at
     26 char, append a '-' and first add 5char of hash.
     This makes sure our usernames are always under 32char.
     """
 
     if len(username) < 26:
-        return username
+        return username.replace(' ', '-')
```
manics, line 19: This results in a clash between `a-b c` and `a b-c`

kamatsuoka: Good point. So maybe it's better to remove any illegal characters and then append a hash of the original name, as we do for long usernames ...

kamatsuoka:
Closing this in favor of https://github.com/jupyterhub/the-littlest-jupyterhub/pull/834

PR #834, opened 2022-09-23
if username contains illegal characters for linux username, remove them and append hash of username

kamatsuoka:
- [x] Add tests


kamatsuoka:
This is for https://github.com/jupyterhub/the-littlest-jupyterhub/issues/832

Issue #830, opened 2022-09-09
ci: our test suite is currently broken and needs to be fixed


consideRatio:
These integration tests were run just now in the main branch: https://github.com/jupyterhub/the-littlest-jupyterhub/actions.

I think the failure has showed up over time due to something changing outside the code in this repo.

manics:
I started investigating in https://github.com/jupyterhub/the-littlest-jupyterhub/pull/839 by adding some additional debug output
- the container fails to start in the Python 3.10 build (which uses Ubuntu 22.04)
- the container starts in the Python 3.6 (Ubuntu 18.04) and 3.9 (Ubuntu 20.04) builds, but `test_nothing_user_writable` fails because the user has group `0` as a supplementary group

pnasrat:
I’m happy to dedicate some time investigating this if that is of use

pnasrat:
I've created a new branch on top of @manics #839 PR with the extra debugging info. The PR I'll create is WIP for debugging.

pnasrat:
Based upon @manics debugging I added my commit to use `os.setgroups([])` to the subprocess to ensure we don't inherit supplemental groups. That fixes the test. However I'm wondering if the actual intent needs to be handled by setting umask in the installer if the desire is not to have conda installed packages like openssl be group writable.

pnasrat:
My fixes to set group and a pinning of Sphinx to fix the docs build seems to do the right thing. I've filed #852 as 18.04 support should go away.

@manics as I built off your PR how is best to handle a cleaned up PR, are the debugging changes something that should be kept

PR #837, opened 2022-11-14
ci: fix deprecation of set-output in github workflows

consideRatio:
- Addresses https://github.com/jupyterhub/team-compass/issues/579 for this repo.

I confirmed that the test failures in this PR are unrelated as they [occurr in the main branch](https://github.com/jupyterhub/the-littlest-jupyterhub/actions/runs/3461480274) as well now. See https://github.com/jupyterhub/the-littlest-jupyterhub/issues/830 about that.


consideRatio:
Self-merging this as a CI only update verified to work already as this PR ran the impacted logic.

PR #839, opened 2022-11-27
Attempt to fix ci

manics:
Investigation and attempt to fix https://github.com/jupyterhub/the-littlest-jupyterhub/issues/830


manics:
See https://github.com/jupyterhub/the-littlest-jupyterhub/pull/851

PR #838, opened 2022-11-27
Release 0.2.0 (JupyterHub 1.*)

manics:
We really should upgrade JupyterHub since it's still on 1.

- https://github.com/jupyterhub/the-littlest-jupyterhub/issues/724

Changing the default `version` in `bootstrap.py` to `latest` should install the most recent non-release tag.

Once we've made this release I think we should upgrade JupyterHub to 3.0.0 https://github.com/jupyterhub/the-littlest-jupyterhub/issues/818, optionally clean-up the docs (including a possible switch to myst), and release `1.0.0`

PR #842, opened 2023-01-11
update: jupyterhub 3, oauthenticator 15, systemdspawner 0.17 (user env: ipywidgets 8)

yuvipanda:
Should upgrade base python version separately

- [x] <s>Add / update documentation</s>
- [ ] <s>Add tests</s>

### DIFF
--- a/tljh/requirements-base.txt
+++ b/tljh/requirements-base.txt
@@ -5,16 +5,14 @@
 #          the requirements-txt-fixer pre-commit hook that sorted them and made
 #          our integration tests fail.
 #
-# For JupyterHub 1.x SQLAlchemy below 2.0.0
-SQLAlchemy<2.0.0
 # JupyterHub + notebook package are base requirements for user environment
-jupyterhub==3.1.*
-notebook==6.5.*
+jupyterhub==3.*
+notebook==6.*
 # Install additional notebook frontends!
 jupyterlab==3.*
-nteract-on-jupyter==2.1.*
+nteract-on-jupyter==2.*
 # nbgitpuller for easily pulling in Git repositories
-nbgitpuller==1.1.*
+nbgitpuller==1.*
 # jupyter-resource-usage to show people how much RAM they are using
 jupyter-resource-usage==0.7.*
 # Most people consider ipywidgets to be part of the core notebook experience
--- a/tljh/requirements-base.txt
+++ b/tljh/requirements-base.txt
@@ -8,17 +8,14 @@
 # For JupyterHub 1.x SQLAlchemy below 2.0.0
 SQLAlchemy<2.0.0
 # JupyterHub + notebook package are base requirements for user environment
-jupyterhub==1.*
-notebook==6.*
+jupyterhub==3.1.*
+notebook==6.5.*
 # Install additional notebook frontends!
 jupyterlab==3.*
-nteract-on-jupyter==2.*
-# Install jupyterlab extensions from PyPI
+nteract-on-jupyter==2.1.*
 # nbgitpuller for easily pulling in Git repositories
-nbgitpuller==1.*
+nbgitpuller==1.1.*
 # jupyter-resource-usage to show people how much RAM they are using
-jupyter-resource-usage==0.6.*
+jupyter-resource-usage==0.7.*
 # Most people consider ipywidgets to be part of the core notebook experience
-ipywidgets==7.*
-# Pin tornado
-tornado>=6.1
+ipywidgets==8.*
--- a/integration-tests/test_hub.py
+++ b/integration-tests/test_hub.py
@@ -346,12 +346,12 @@ async def test_idle_server_culled():
             )
         ).wait()
     )
-    # Check every 10s for idle servers to cull
+    # Check every 5s for idle servers to cull
     assert (
         0
         == await (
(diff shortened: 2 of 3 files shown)


consideRatio:
I'd love to see TLJH get more up to date, and this is part of it - upgrading jupyterhub from 1 to 3!

I'm currently uncomfortable making a package upgrade as we are behind on maintenance work.

I'm open to going for a change like this to get some momentum in repo maintenance efforts, but I'd really would like the CI system to work again first.

Overall I'd be a lot more comfortable maintaining this repo if we got:
- ci system to work
- a release process defined and version a 1.0.0 release made
- ci system to have an upgrade test

@manics has done a lot of work towards this already!

tljh/requirements-base.txt
```diff
@@ -8,17 +8,14 @@
...
+jupyterhub==3.1.*
+notebook==6.5.*
 # Install additional notebook frontends!
 jupyterlab==3.*
-nteract-on-jupyter==2.*
-# Install jupyterlab extensions from PyPI
+nteract-on-jupyter==2.1.*
 # nbgitpuller for easily pulling in Git repositories
-nbgitpuller==1.*
+nbgitpuller==1.1.*
```
consideRatio, line 17: These are pinned to the minor version, I think we should stick with major versions here as well.

I can't say if that was an update part of the change you pushed now though Min, so now this is a bit mixxed up.

minrk: I didn't change anything, I just rebased so it would run with the tests. I'm fine with shifting to just major versions.

minrk: Now this only pins major versions, which is probably right when we aren't doing proper frozen environments

minrk:
Oop, realized this is going to have to wait for #856 because Python 3.6 support has been dropped by lots of packages. Indeed, so has 3.7, so I think we should require debian 11, which has been stable for 2 years already.

tljh/requirements-base.txt
```diff
@@ -5,20 +5,15 @@
...
 #
-# For JupyterHub 1.x SQLAlchemy below 2.0.0
-SQLAlchemy<2.0.0
 # JupyterHub + notebook package are base requirements for user environment
-jupyterhub==1.*
+jupyterhub==3.*
 notebook==6.*
 # Install additional notebook frontends!
 jupyterlab==3.*
 nteract-on-jupyter==2.*
```
consideRatio, line 13: Last release, 2019, this is deprecated according to: https://github.com/nteract/nteract/issues/5554#issuecomment-852434391

Should we remove it now as well?

minrk: Removed.

yuvipanda:
Removed nteract_on_jupyter from repo2docker too https://github.com/jupyterhub/repo2docker/pull/1259

minrk:
I've got this working (needs a tiny fix in hubtraf), but now GitHub's down so I can't push anywhere. Will come back to it tomorrow, if GitHub isn't prompt coming back.

minrk:
I've got this working (needs a tiny fix in hubtraf), but now GitHub's down so I can't push anywhere. Will come back to it tomorrow, if GitHub isn't prompt coming back.

minrk:
All green! Had to update a couple of tests, but nothing of consequence (that is tested) appears to have broken.

yuvipanda:
yay, thank you so much, @minrk!

Given that:

1. The CI system is now green, in this PR
2. https://github.com/jupyterhub/the-littlest-jupyterhub/pull/838 was merged
3. @minrk did most of the work even though I opened the PR

I feel this is good to merge :D

yuvipanda:
I know @consideRatio is out this week. If this isn't merged by next week and there are no objections, I'll merge it. I'm happy for others to merge it before that too, of course!

consideRatio (approved):
LGTM, nice work!

@manics has done a lot of thinking about handling releases for this distribution - which is a bit of an edge case compared to other projects due to the bootstrap script etc.

Simon do you think this change is fine to merge right away, or is there something we ought to get first?

minrk:
I think it's important to think about (and document after #863) upgrade expectations, and in what situations the installer will run, then we can make sure our commitments are tested and do what we intend. I've implemented what I described above, which I think is reasonable, but of course there are other reasonable choices to make, as well.

manics:
The installer _should_ install the last tagged version.... if you want to be absolutely sure then do an install and check the version is `0.2.0` and not:
https://github.com/jupyterhub/the-littlest-jupyterhub/blob/adbd41dcb023aff944a6b66c4b912b1ed382c2f9/setup.py#L5

Is the plan to add up upgrade test in a future PR before tagging 1.0.0?

manics:
I've checked, `curl -L https://tljh.jupyter.org/bootstrap.py | sudo -E python3 - --admin admin` installs 0.2.0, `curl -L https://tljh.jupyter.org/bootstrap.py | sudo -E python3 - --admin admin --version main` installs the main branch

consideRatio:
> Is the plan to add up upgrade test in a future PR before tagging 1.0.0?

I opened https://github.com/jupyterhub/the-littlest-jupyterhub/issues/868 about this, it doesn't look that hard to get done so I think yes that is worth getting done.

PR #855, opened 2023-03-20
added `remove_named_servers` setting for jupyterhub-idle-culler

stevejpurves:
If you are using TLJH with named servers, it can be a requirement that the server is removed when culled. This PR is to expose the underlying IdleCuller option, allowing it to be set via `tljh-config`.

- [x] Add tests

### DIFF
--- a/docs/topic/idle-culler.rst
+++ b/docs/topic/idle-culler.rst
@@ -44,6 +44,12 @@ the users will not be culled alongside their notebooks and will continue to exis
 
 	services.cull.users = False
 
+If named servers are in use, remove the server after stopping it.
+
+.. code-block:: python
+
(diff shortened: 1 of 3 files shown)


stevejpurves:
I noted some issues with the test suite while getting this to go green. It seems like he mock config can leak between tests meaning it was not possible to add a test for `remove_named_servers` independently of `users`. 

Also I added a test case where potentially the existing test case for services could be extended instead -- let me know if I should change that.

minrk (approved):

consideRatio (approved):
This PR has merge conflicts that needs to be manually resolved, but this LGTM and can be merged when those are resolved!

Thank you @stevejpurves for your work on this!!! :heart: :tada: :sunflower:

PR #856, opened 2023-03-21
require ubuntu 20.04, test on debian 11, require Python 3.8

### LINKED ISSUE #813: Add support for Ubuntu 22.04
### Proposed change
Add support for Ubuntu 22.04

### Alternative options
Do nothing

People using the latest Ubuntu LTS

22.04 doesn't work according to this report: https://discourse.jupyter.org/t/tljh-failing-on-ubuntu-22-04-lts/14097

20.04 seems to work but is not formally documented, see https://github.com/jupyterhub/the-littlest-jupyterhub/issues/720
As part of adding 22.04 we may drop support for 18.04 (as mentioned in https://github.com/jupyterhub/the-littlest-jupyterhub/issues/720)

We should probably also add proper releases as described in https://github.com/jupyterhub/the-littlest-jupyterhub/issues/724 if supporting 22.04 requires non-trivial changes.

### LINKED ISSUE #852: Remove support for Ubuntu 18.04 (deprecated environment)
Got below 18.04 brownout message in the test runs for a PR. As 18.04 is EOL (with only ESM support available). 

Support for 18.04 should be removed from tljh.

This is a scheduled Ubuntu-18.04 brownout. The Ubuntu-18.04 environment is deprecated and will be removed on April 1st, 2023. For more details, see https://github.com/actions/runner-images/issues/6002

minrk:
- [x] Add tests

I assume this isn't going to pass right away...

I removed the runs-on and Python version matrix entries for the integration tests, since these values only exercise the host system, which isn't where the tests run. The tests all actually run in containers. There's no real reason for variation there, that I can see.

closes #852 
closes #813 (already seems to work fine on 22.04, we have tests running, but this makes it clear in the docs)


docs/index.rst
```diff
@@ -17,7 +17,7 @@ might still make breaking changes that have no clear upgrade pathway.
 Installation
 ============
 
-The Littlest JupyterHub (TLJH) can run on any server that is running **Ubuntu 18.04** or **Ubuntu 20.04** on a amd64 or arm64 CPU architecture. Earlier versions of Ubuntu are not supported.
+The Littlest JupyterHub (TLJH) can run on any server that is running **Debian 10** or **Ubuntu 20.04** on a amd64 or arm64 CPU architecture. Earlier versions of Ubuntu and Debian are not supported, as are other Linux distributions.
```
consideRatio, line 20: We test with Ubuntu 22.04 as well, so that makes sense to include still, and I figure we look to support Debian 11 as well.

minrk: Yup, sometimes it says `>=`, but here it implies exact. What do we want to claim? Do we want to claim only exactly `20.04, 22.04`, or `>=20.04`?

consideRatio: I figure that we aim to support LTS releases of Ubuntu 20.04 and Debian 11 or higher, and currently we test for Ubuntu 20.04, 22.04, and debian 11.

minrk: added:

> We aim to support 'stable' and Long-Term Support (LTS) versions.
Newer versions are likely to work with little or no adjustment, but these are the officially supported and tested versions.

to try to communicate that 24.04 _may_ work, but if it's not in the list above, it's not 'supported'

consideRatio: Looked at the commit, I think it was great!!

consideRatio (approved):
LGTM! I think the title could benefit from noting that we require python 3.8 also!

Since this is a breaking change I'm pinging @manics for another review, but I also want this to land quickly. So I think we should go for a merge sooner rather than later if Simon doesn't find time to checkin on this.

minrk:
All set from my side. I think next steps are:

- #842 to update packages
- update mambaforge version, base environment

consideRatio:
@minrk I see that this PR includes the commits from #843 - should we merge this and rebase on that or similar to help with attribution?

minrk:
By including the commits, attribution should be properly assigned.

consideRatio:
👍 and it will close by merging this PR because the commits get merged i assume! Sounds good!

minrk:
That's my understanding. Ready to merge, or do we want another set of eyes on this?

bootstrap/bootstrap.py
```diff
@@ -199,22 +204,22 @@ def get_os_release_variable(key):
...
         print("The Littlest JupyterHub currently supports Ubuntu or Debian Linux only")
         sys.exit(1)
-    elif distro == "ubuntu" and float(version) < 18.04:
-        print("The Littlest JupyterHub requires Ubuntu 18.04 or higher")
+    elif distro == "ubuntu" and _parse_version(version) < (20, 4):
+        print("The Littlest JupyterHub requires Ubuntu 20.04 or higher")
         sys.exit(1)
-    elif distro == "debian" and float(version) < 10:
+    elif distro == "debian" and _parse_version(version) < (11,):
         print("The Littlest JupyterHub requires Debian 10 or higher")
```
manics, line 217:
```suggestion
         print("The Littlest JupyterHub requires Debian 11 or higher")
```

bootstrap/bootstrap.py
```diff
@@ -9,11 +9,10 @@
 
 Constraints:
 
-    - The entire script should be compatible with Python 3.6, which is the on
-      Ubuntu 18.04+.
-    - The script should parse in Python 3.5 as we print error messages for using
-      Ubuntu 16.04+ which comes with Python 3.5 by default. This means no
-      f-strings can be used.
+    - The entire script should be compatible with Python 3.8, which is the on
```
manics, line 12:
```suggestion
    - The entire script should be compatible with Python 3.8, which is the default on
```

docs/index.rst
```diff
@@ -17,7 +17,10 @@ might still make breaking changes that have no clear upgrade pathway.
 Installation
 ============
 
-The Littlest JupyterHub (TLJH) can run on any server that is running **Ubuntu 18.04** or **Ubuntu 20.04** on a amd64 or arm64 CPU architecture. Earlier versions of Ubuntu are not supported.
+The Littlest JupyterHub (TLJH) can run on any server that is running **Debian 11** or **Ubuntu 20.04** or **22.04** on an amd64 or arm64 CPU architecture.
+We aim to support 'stable' and Long-Term Support (LTS) versions.
+Newer versions are likely to work with little or no adjustment, but these are the officially supported and tested versions.
```
manics, line 22:
```suggestion
Newer versions are likely to work with little or no adjustment, but these are not officially supported or tested.
```

docs/install/azure.rst
```diff
@@ -13,16 +13,16 @@ users and a user environment with packages you want to be installed running on
...
     Check it out at `https://github.com/trallard/TLJH-azure-button <https://github.com/trallard/TLJH-azure-button>`_.
 
 Prerequisites
 ==============
 
-* A Microsoft Azure account. 
+* A Microsoft Azure account.
 
-* To get started you can get a free account which  includes 150 dollars worth of Azure credits (`get a free account here <https://azure.microsoft.com/en-us/free//?wt.mc_id=TLJH-github-taallard>`_) 
+* To get started you can get a free account which  includes 150 dollars worth of Azure credits (`get a free account here <https://azure.microsoft.com/en-us/free//?wt.mc_id=TLJH-github-taallard>`_)
```
manics, line 25: Do we need the tracking(?) identifier here and in the other Azure links?

minrk: I'm not sure, @trallard can you comment on the campaign ID in the azure links (added in #342)?

consideRatio: I opened https://github.com/jupyterhub/the-littlest-jupyterhub/issues/860 to track this so that we can move onwards with this PR

PR #858, opened 2023-03-21
Update base user environment to mambaforge 23.1.0-1 (Python 3.10)

### LINKED ISSUE #864: install error:  Downloading & setting up user environment, Exception sha256sum hash mismatch!
Hi，I have a problem when installing TLJH. Could you tell me how to install TLJH ? 
install TLJH with : 
    curl -L https://tljh.jupyter.org/bootstrap.py | sudo -E python3 - --admin  admin

log:
Existing TLJH installation not detected, installing...
Setting up hub environment...
Installing Python, venv, pip, and git via apt-get...
Setting up virtual environment at /opt/tljh/hub
Upgrading pip...
Installing TLJH installer...
Running TLJH installer...
Setting up admin users
Granting passwordless sudo to JupyterHub admins...
Setting up user environment...
Downloading & setting up user environment...

  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/opt/tljh/hub/lib/python3.10/site-packages/tljh/installer.py", line 478, in <module>
    main()
  File "/opt/tljh/hub/lib/python3.10/site-packages/tljh/installer.py", line 452, in main
    ensure_user_environment(args.user_requirements_txt_url)
  File "/opt/tljh/hub/lib/python3.10/site-packages/tljh/installer.py", line 197, in ensure_user_environment
    with conda.download_miniconda_installer(
  File "/usr/lib/python3.10/contextlib.py", line 135, in __enter__
    return next(self.gen)
  File "/opt/tljh/hub/lib/python3.10/site-packages/tljh/conda.py", line 64, in download_miniconda_installer
    raise Exception("sha256sum hash mismatch! Downloaded file corrupted")
Exception: sha256sum hash mismatch! Downloaded file corrupted

minrk:
shifts some duplicated code into utility functions and constants

This significantly updates both conda and mamba. It also updates the default python to 3.10

It still doesn't affect the user environment if it hasn't been updated

closes #864


tljh/installer.py
```diff
@@ -154,58 +156,89 @@ def ensure_usergroups():
...
+                found_conda = True
+                conda_version, mamba_version = conda_mamba_version
+                break
+
+    if not found_conda:
+        if os.path.exists(USER_ENV_PREFIX):
+            logger.warning(
+                f"Found prefix at {USER_ENV_PREFIX}, but too old or missing conda/mamba ({have_versions}). Rebuilding env from scratch!!"
+            )
+            # FIXME: should this fail? I'm not sure how destructive it is
```
manics, line 226: We should check this before merging! If in doubt I think failing and telling the admin to do X is OK, especially since this will be a breaking release, and we'll need to update jupyterhub-singleuser anyway. If it's easier I think we could say a new user env will always be created (and rename the old one)

minrk: Test added. Re-running the installer does not wipe out the env, but it will upgrade many packages significantly. I'm sure there are cases where this can fail, but I think the behavior should be sensible and not too destructive.

I think it's an open question if/whether we should update conda/mamba from older versions, but at least this isn't changed by this PR, other than moving the 'way too old' bar up to 2021.

manics: If the situation is no worse than before then let's not worry about it in this PR

minrk:
Still trying to understand the failure when upgrading from too-old conda 4.7. Especially since when I run the test in a Dockerfile, it succeeds!

minrk:
I'm thinking that probably the right thing to do is to fail explicitly, rather than trying to upgrade.

minrk:
I'm struggling a bit with this one, particularly what upgrade situations should be handled, and what assumptions should be made.

Past equivalent changes in the _initial_ tljh user env:

- early tljh, the user env was Python 3.6, via miniconda 4.5.4
- #433 updated the default user env to Python 3.7 via miniconda 4.7.10
- #697 updated to Python 3.9 via mambaforge 4.10.3-7

And the code _suggests_ that it is attempting to prevent upgrades across those versions, but after writing some tests, I don't believe this is likely to work as intended, with the possible exception of upgrading from the latest version to the latest version.

I _believe_ the changes in #697 prevented upgrades from earlier commits, because they switched `ensure_conda` to calling mamba, but then use that function to install mamba itself. If that was run on an earlier tljh install, it would fail with `mamba: command not found` when trying to install mamba itself. I believe that means that no tljh installed before #697 has ever been upgraded past #697 (at least not without manual intervention to install mamba, first).

Further, I've noticed that `auto_update_conda` is not set to False, so if the user ever ran a `conda install` command, conda would have been updated, and the pinning logic in the next run to installer.py would not do what was expected. For example, a Python 3.7 env could easily have been upgraded to conda 4.14 or later, which would then get _downgraded_ in the install stage to 4.10.3,  not the expected 4.5.8.

So I have these questions about the user environment:

- does tljh assume it 'owns' the user env directory, i.e. that it was created by tljh and can be destroyed by tljh (i.e. ignore the case of the directory existing with arbitrary unrecognized contents?)
- what user upgrades can be expected in the user env? Updates to Python version? Updates to conda version? It seems that all of these are allowed and supported.
- do we really want to pin _down_ the conda/mamba versions?

It seems that the goal of the if/else branches we have are to prevent breaking environments by upgrading Python itself too far, but the conda version is used as a _proxy_ for this, when it isn't actually reliable to do so.

So I think what makes the most sense is the following behavior:

1. specify a minimum supported conda and/or mamba
2. if it's not present or too old, run a generic `conda update conda [mamba]` (no version pin, or >=lower bound)
3. specify a minimum supported Python version and fail if it's too old, refusing to upgrade


This will result in the following different behavior:

1. conda/mamba will be up-to-date, not pinned to a point in time
2. if conda has been upgraded by users, it will no longer be downgraded
3. if an unsupported environment is detected, fail until user intervention (e.g. backup and clear the environment to start fresh), rather than clobber by re-running the latest mambaforge installer

How does that sound?

manics:
IIRC in #697 I based my changes on the existing code which was also unclear to me!

Regarding upgrades I think your comment about when TLJH should make changes to the user env is key. Any automated upgrade of a user environment package risks "breaking" the user environment, since it may be incompatible with the users' notebooks/code, or it may
transitively upgrade another package which is incompatible.

We can't completely avoid updates though, since JupyterHub major releases require a singleuser upgrade.

Now that this repo follows semver could we simplify things?
- Never make changes to the user environment in a minor/patch release
- For major releases backup the user environment, then upgrade all required TLJH packages (including conda/mamba)

This means we'd get rid of the multiple version handling, and makes it clear to admins and users when the user environment will change. It also helps avoid some inconsistency e.g. if TLJH needs to update an existing package do we bump them to the minimum required version (minimising the difference from the existing installed versions), or a more recent version as would be installed in a new installation of TLJH?

minrk:
OK, I think this behavior is ready for review. I'd like to update docs, but I think that needs to wait for #863 to avoid conflict.

The new behavior:

- Python version is checked directly. Minimum Python for the user env is 3.9 (the version in 0.2, since 2021). If this requirement is not met, the installer aborts (change from earlier behavior with user envs on old Python, installed prior to 2021)
- conda/mamba only have lower bounds, not exact pins. They include the _previous_ versions (i.e. will not need updates when upgrading from 0.2). If conda or mamba is out of date or mamba is missing, the packages are upgraded (unbounded). That means these packages will no longer be downgraded during tljh install. But it also means that _outdated_ envs will get updated to 'current' conda/mamba instead of a pinned version. This matches the existing behavior if the user has ever run `conda install`, since we leave `auto_update_conda` unset.
- if the directory exists, is not empty, and does not contain a conda installation, install will abort instead of attempting to clobber the non-environment directory

I tried running the unbounded upgrade for Python 3.7, and it took a very long time, I think because of conflicts on openssl since the versions are so out of date. This is part of why I set the minimum version to what's in the most recent mambaforge install (3.9).

consideRatio:
> The new behavior:
> 
> * Python version is checked directly. Minimum Python for the user env is 3.9 (the version in 0.2, since 2021). If this requirement is not met, the installer aborts (change from earlier behavior with user envs on old Python, installed prior to 2021)
> * conda/mamba only have lower bounds, not exact pins. They include the _previous_ versions (i.e. will not need updates when upgrading from 0.2). If conda or mamba is out of date or mamba is missing, the packages are upgraded (unbounded). That means these packages will no longer be downgraded during tljh install. But it also means that _outdated_ envs will get updated to 'current' conda/mamba instead of a pinned version. This matches the existing behavior if the user has ever run `conda install`, since we leave `auto_update_conda` unset.
> * if the directory exists, is not empty, and does not contain a conda installation, install will abort instead of attempting to clobber the non-environment directory

I appreciate this a lot!!!

Related but off topic, I think we should minimize what is installed in the user environment, and have opened #872 about this.

tljh/installer.py
```diff
@@ -154,60 +156,115 @@ def ensure_usergroups():
...
 
 
+# Install mambaforge using an installer from
+# https://github.com/conda-forge/miniforge/releases
+MAMBAFORGE_VERSION = "22.11.1-4"
+# sha256 checksums
+MAMBAFORGE_CHECKSUMS = {
+    "aarch64": "96191001f27e0cc76612d4498d34f9f656d8a7dddee44617159e42558651479c",
+    "x86_64": "16c7d256de783ceeb39970e675efa4a8eb830dcbb83187f1197abfea0bf07d30",
+}
```
consideRatio, line 166:
```suggestion
MAMBAFORGE_VERSION = "23.1.0-1"
# sha256 checksums
MAMBAFORGE_CHECKSUMS = {
    "aarch64": "d9d89c9e349369702171008d9ee7c5ce80ed420e5af60bd150a3db4bf674443a",
    "x86_64": "cfb16c47dc2d115c8b114280aa605e322173f029fdb847a45348bf4bd23c62ab",
}
```

tests/test_installer.py
```diff
@@ -36,3 +43,194 @@ def test_ensure_admins(tljh_dir, admins, expected_config):
...
+    "distro, distro_version, expected_versions",
+    [
+        # No previous install, start fresh
+        (
+            None,
+            None,
+            {
+                "python": "3.10.*",
+                "conda": "22.11.1",
+                "mamba": "1.1.0",
```
consideRatio, line 121:
```suggestion
                "conda": "23.1.0",
                "mamba": "1.4.1",
```

tests/test_installer.py
```diff
@@ -36,3 +43,194 @@ def test_ensure_admins(tljh_dir, admins, expected_config):
...
+            },
+        ),
+        # previous install, 1.0
+        (
+            "mambaforge",
+            "22.11.1-4",
+            {
+                "python": "3.10.*",
+                "conda": "22.11.1",
+                "mamba": "1.1.0",
```
consideRatio, line 131:
```suggestion
                "conda": "23.1.0",
                "mamba": "1.4.1",
```

tests/test_installer.py
```diff
@@ -36,3 +43,194 @@ def test_ensure_admins(tljh_dir, admins, expected_config):
...
+        ),
+        # simulate missing mamba
+        # will be installed but not pinned
+        # to avoid conflicts
+        (
+            "miniforge",
+            "4.10.3-7",
+            {
+                "conda": "4.10.3",
+                "mamba": ">=1.1.0",
```
consideRatio, line 152:
```suggestion
                "mamba": ">=1.4.1",
```

consideRatio (approved):
Amazing work on this @minrk and @manics!!!

I suggested using the currently latest available mambaforge installation (23.1.0-1) that associated to python 3.10, mamba 1.4.1, and conda 23.1.0

- [x] Update PR description also if committing the code suggestions to update to latest mambaforge

tljh/installer.py
consideRatio: This looks great to me now, marking this as resolved.

consideRatio:
A test failure relates to deprecated use of `codecov` PyPI package, resolved by #876 that also adds upgrade tests relevant to use to test this PR.

tests/test_installer.py
```diff
@@ -36,3 +43,194 @@ def test_ensure_admins(tljh_dir, admins, expected_config):
...
+            {
+                "python": "3.10.*",
+                "conda": "23.1.0",
+                "mamba": "1.4.1",
+            },
+        ),
+        # previous install, 1.0
+        (
+            "mambaforge",
+            "22.11.1-4",
```
consideRatio, line 127: I forgot this in the patch I committed @minrk!

```suggestion
            "23.1.0-1",
```

consideRatio: Since tests takes quite a wile, I figure I'll commit this to reduce context switching.

tests/test_installer.py
```diff
@@ -149,7 +149,7 @@ def _specifier(version):
             "4.10.3-7",
             {
                 "conda": "4.10.3",
-                "mamba": ">=1.4.1",
+                "mamba": ">=1.1.0",
```
consideRatio, line 152: This was installed fresh but still got stuck at 1.1.0, I guess it was held back for some reason. I find digging into why a conda-forge package is held back during install/upgrade is soo tricky, and I failed with my preliminary investigations looking into the mamba recipie for example.

consideRatio: > it may not upgrade all the way to 1.4 if other packages are pinned in the base env

:+1:

tests/test_installer.py
```diff
@@ -36,3 +43,194 @@ def test_ensure_admins(tljh_dir, admins, expected_config):
...
+        ),
+        # simulate missing mamba
+        # will be installed but not pinned
+        # to avoid conflicts
+        (
+            "miniforge",
+            "4.10.3-7",
+            {
+                "conda": "4.10.3",
+                "mamba": ">=1.4.1",
```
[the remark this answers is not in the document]
minrk: `>=` because it's an unbounded install. It will and should increase over time. When I run this on arm64 today, it picks 1.4.2. When I run it on x86_64, it picks 1.1.0. I presume this is related to the pinning of something like openssl or libc++ in the base environment, causing the install to reject the latest build. But it's not really important to us that it picks the _latest_ version, only that it's 

1. present, and
2. recent _enough_
 

our needs are minimal at this point, so just about any version is recent enough. But if we do start to have a reason to require a minimal version, we can update the test to verify it.

consideRatio:
Super happy about this work being done @minrk!!! Thank you!!

PR #862, opened 2023-03-24
avoid registering duplicate log handlers

minrk:
init_logging is called many times in test_config, which has been registering numerous duplicate log handlers, attached to stderr that's closed between tests


minrk:
This is causing lots of noise and "ValueError: I/O operation on closed file.", e.g. [here](https://github.com/jupyterhub/the-littlest-jupyterhub/actions/runs/4502308353/jobs/7923943281?pr=858#step:8:759)

Issue #885, opened 2023-05-09
ci: broken without change in repo


consideRatio:
I've not been able to figure out what is wrong in our CI system, but I've condluded it has started failing without changes to this repo after April 17 09:00 UTC sometime, before April 24.

### When failure started

[3 weeks ago - failure](https://github.com/jupyterhub/the-littlest-jupyterhub/actions/runs/4719827988)
[3 weeks ago - success](https://github.com/jupyterhub/the-littlest-jupyterhub/actions/runs/4719827988)

![image](https://user-images.githubusercontent.com/3837114/237032075-7289c7bc-3ddb-440c-9eaf-ef9f10e1f5a6.png)

### About failure

I really don't know what the failure is about, and I don't know how to clearly see the jupyterhub logs associated with the failure.

### Related

Test failure indicates this could be related to hubtraf and XSRF matters, I opened https://github.com/yuvipanda/hubtraf/issues/36 but that may be a distraction. I really can't say what this is about!

minrk:
I think this is due to a lack of the xsrf token in the login request. But that should mean jupyterhub 4 is used, and our tests shouldn't be running with hub 4, should they?

PR #886, opened 2023-05-11
add integration test for hub version

minrk:
make sure the hub has the version we expect

I think #885 is caused by jupyterhub 4, but I don't yet understand how/why we are getting that


minrk:
Seems I was wrong about JupyterHub 4 being the culprit. I'm a bit confused because I thought I fixed the xsrf issue in https://github.com/yuvipanda/hubtraf/pull/35 (March 27). I don't know what changed since April 17, which was the successful build.

Issue #873, opened 2023-04-09
Planning for release 1.0.0


consideRatio:
This is a placeholder PR to track work towards the 1.0.0 release, maintainers please feel free to edit this!

### Blocking issues/PRs

- #858
- #868
- #880 
- #901
- #915 
- #899 
- #918 
- #920 
- https://github.com/jupyterhub/oauthenticator/issues/585
- #924
- #888 
- [ ] There be notes to link to from the changelog on how to make an upgrade and what to think about when doing one
  - #81 

### Relevant but maybe not blocking issues/PRs

- #872
- #874
- https://github.com/jupyterhub/the-littlest-jupyterhub/issues/909

### Helpful notes for creating the changelog

> > Can we upgrade straight from jupyterhub 1 to jupyterhub 4?
> 
> It's tested in the jupyterhub suite and should be supported. It's a disruptive upgrade (the 1->2 transition means all user servers must be restarted and any API-issued tokens will be revoked), but for most tljh use cases, the restart-everything should be the only issue folks need to deal with.

consideRatio:
I moved #872 to blocking as I don't want `pip install --upgrade` of requirements-base.txt to cause issues in the user environment that is unrelated to making the jupyterhub itself function.

nsurleraux-railnova:
Hi @consideRatio,

First, I would like to say thank you for the amazing work on TLJH. Without wanting to be pushy, is there an ETA on 1.0.0 release? We would need to have JupyterHub>=4.0 for JupyterLab extensions to be picked up correctly (see https://github.com/jupyter-server/jupyter-scheduler/issues/370). If there are any known workaround in the meantime, I would highly appreciate it.

Thanks !

consideRatio:
@nsurleraux-railnova a pull request to add the changelog for 1.0.0 is opened and ready for review, you are most welcome to provide feedback on the text if you want!

consideRatio:
A beta release is out, check the changelog for details. At this point what remains is feedback about using the beta and documentation on how to upgrade in general.

The changelog is available at https://tljh.jupyter.org/en/latest/reference/changelog.html

godber:
Will this 1.0 release include JupyterLab 4 and Jupyter Notebook 7, both of which recently released.

BTW, I was able to upgrade from the current latest to the `1.0.0.b1` on a small install with no problem, I had to restart the upstream reverse proxy for some reason, but it was fine after that.

jtpio:
[...]

Issue #81, opened 2018-07-28
Document upgrade strategy


yuvipanda:
People should be able to upgrade 1 minor version of TLJH installations.

- [x] Decide on & document our upgrade strategy
  - Users are to run bootstrap.py script from tljh's main branch from where they can specify the version they wish to install/upgrade to.
  - The Hub envs python packages gets upgraded, but the user env is touched as little as possible
- [x] Document various components & how they would react to upgrades
  - The changelog in #888 mentions this
- [x] Write upgrade documentation
  - Resolved by #932
- [ ] Document what to do if we want a 'wipe everything and start over, but preserve user data' upgrade
  - Still needed as of 2023-08-10, this issue is closed by #932 without addressing this

---

## Related
- #724
- #204

yuvipanda:
I think the general upgrade strategy is to run the installer again, and that 'ensures' everything to where it should be. This means we need a way in the installer to:

1. Tell what the current version is
2. Refuse to upgrade if it is unsupported
3. Upgrade various components as atomically as possible
4. Always have option of 'blow everything away and re-install it' without losing any user state.

choldgraf:
Plan is to include this in v0.2, since we don't necessarily want to "freeze" an upgrade pathway yet until we see more usecases out of TLJH.

nsurleraux-railnova:
If anything, I would love to have my installation backed up so that if an upgrade borks it, we can always restore it to the latest working state. (It could be automated or part of the documented upgrade process).

In the [requirements-user-env-extras.txt](https://github.com/jupyterhub/the-littlest-jupyterhub/blob/main/tljh/requirements-user-env-extras.txt) file, it is clearly written that TLJH **won't** upgrade those libraries if TLJH is upgraded. The documentation should explain why it won't be upgraded (as users can upgrade/downgrade them). I think it would be reasonable in the upgrade documentation to have a table with the versions of the TLJH against the recommended versions of the libraries in the `user` env.

PR #888, opened 2023-05-13
Add changelog for 1.0.0b1

consideRatio:
I propose we make a 1.0.0b1 release ~today (July 6th 2023)! I know there are at least a few users very interested in getting a release out based on comments in issues. I'm going on vacation next week, and really hope to have a beta release out by then. I have a lot of other things I need to do by then as well, so I'm extremely thankful for all help to get this pre-release out.

## Action points

- This changelog includes the PR #924 that should get merged before the 1.0.0b1 pre-release

## Action points that maybe doesn't block a 1.0.0b1 release

- [ ] There should ideally be notes to link to from the changelog on how to make an upgrade and what to think about when doing one
  - #81 

## Related

- #873

## Changelog preview

See it at [RTD PR build[(https://the-littlest-jupyterhub--888.org.readthedocs.build/en/888/reference/changelog.html), here is a screenshot.

image


consideRatio:
Hmmm, I think maybe we want:
- #81 
- Link to the result of #81 from the changelog or write in it that one is expected to run the bootstrap.py script as fetched from the internet.

nsurleraux-railnova (approved):
Great work! It still needs documentation about the upgrading process, especially about upgrading user env libs and making a backup if the upgrade borks the user env.

nsurleraux-railnova:
> When you write "about upgrading user env libs", what are you thinking about more specifically? I'm thinking that this upgrade should not need to be associated with upgrading the user env in any way, and that users should manage that independently on this upgrade being made.

In your changelog, you are including a table named`Dependency changes in the user environment`, which may lead to think that they are part of the automated upgrade process. But the code specifies that they won't be upgraded. I would like to see in the documentation of the upgrade process:
- Backup your `/opt/tljh/user` folder.
- Run `/opt/tljh/user/bin/pip install jupyterhub==4.x.x`

Or something similar...

Another option could be to have an option in the upgrade process (maybe `--upgrade-user-env`) that will reinstall the user env dependencies from https://github.com/jupyterhub/the-littlest-jupyterhub/blob/main/tljh/requirements-user-env-extras.txt. It should be pretty clear in the documentation that a backup is always necessary or the upgrade process should automatically backup the user env...

nsurleraux-railnova:
Hi @consideRatio,

Will the first release of 1.0.0 be a beta version since you changed the version tag to 1.0.0b1? What motived this change of direction?

consideRatio:
@nsurleraux-railnova yes making an initial beta release is the current proposal. I made a proposal for 1.0.0 before because I didn't understand it was technically possible for this project to make make a beta release until yesterday via https://github.com/jupyterhub/the-littlest-jupyterhub/issues/909#issuecomment-1621892582.

nsurleraux-railnova:
Ok! Should we then consider this beta as not production ready?

minrk (approved):

consideRatio:
Thanks @minrk!!

PR #890, opened 2023-05-13
Only upgrade jupyterhub in user env when upgrading tljh

consideRatio:


tljh/installer.py
```diff
@@ -267,9 +271,15 @@ def ensure_user_environment(user_requirements_txt_file):
 
     conda.ensure_pip_requirements(
         USER_ENV_PREFIX,
-        os.path.join(HERE, "requirements-base.txt"),
+        os.path.join(HERE, "requirements-user-env.txt"),
         upgrade=True,
```
minrk, line 275: Maybe only specify `--upgrade` if it's a fresh install?

```suggestion
        upgrade=is_fresh_install,
```

Differences:

1. if not upgrade, if the requirement is satisfied no changes will occur, even if there are newer versions that satisfy the requirements (e.g. have 4.0.0, 4.0.2 is out). 
2. if upgrade is specified, jupyterhub _and all its dependencies_ (e.g. sqlalchemy) will be upgraded to their latest version, regardless of whether the env already satisfies the requirements.

I'm genuinely not sure what's the best call here - refreshing jupyterhub, etc. may well be good in many cases (e.g. getting patch releases).

An example: `pip install --upgrade jupyterhub==1.*` will take a working env that satisfies all requirements and upgrade to a _broken_ env by installing sqlalchemy 2.0.

consideRatio: I'm reflecting on:
- Use of `==4.*` vs `>=4.0.0,<5` in the jupyterhub specification, thinking we should go for the latter to let a new TLJH release force an upgrade from 4.0.0 to 4.0.1 or similar by declaring `>=4.0.1,<5` if we don't do `--upgrade`.
- If `pip install --upgrade-strategy` is relevant.

  It seems that `pip install --upgrade` with pip 23.1.2 has an upgrade strategy of `only-if-needed` by default, which indicates to me that sqlalchemy would remain at an older version if some version of jupyterhub was installed and the new version of jupyterhub installed didn't declare an sqlalchemy constraint that didn't work.
  ```
  -U, --upgrade               Upgrade all specified packages to the newest
                              available version. The handling of dependencies
                              depends on the upgrade-strategy used.
  --upgrade-strategy <upgrade_strategy>
                              Determines how dependency upgrading should be
                              handled [default: only-if-needed]. "eager" -
                              dependencies are upgraded regardless of whether
                              the currently installed version satisfies the
                              requirements of the upgraded package(s). "only-
                              if-needed" -  are upgraded only when they do not
                              satisfy the requirements of the upgraded
                              package(s).
  ```

I'm open to not declaring `--upgrade`, but I think that comes with some additional maintenance work.

If its a fresh install, the `--upgrade` flag doesn't matter so we can simplify it to this I think.

```suggestion
```

tljh/installer.py
```diff
@@ -267,9 +271,15 @@ def ensure_user_environment(user_requirements_txt_file):
...
     )
+    if is_fresh_install:
+        conda.ensure_pip_requirements(
+            USER_ENV_PREFIX,
+            os.path.join(HERE, "requirements-user-env-extras.txt"),
+            upgrade=True,
```
consideRatio, line 281:
```suggestion
```

Its a fresh install, so `upgrade` doesn't matter.

tljh/installer.py
manics: I think `--upgrade only-if-needed` makes sense (assuming it works intuitively..... we should test it to be sure!)

tljh/requirements-user-env.txt
```diff
@@ -0,0 +1,7 @@
+# When tljh.installer runs, the users' environment as typically found in
+# /opt/tljh/user, is installed or upgraded with these packages.
+#
+# Whats listed here should be only the uncoditional requirements that we have on
```
manics, line 4:
```suggestion
# Whats listed here should be only the unconditional requirements that we have in
```

tljh/installer.py
minrk: I forgot that `only-if-needed` is the default now! So the only difference is for jupyterhub itself, because any _requested_ packages are upgraded to the latest version within the constraints. So the question is: do we want to always upgrade jupyterhub to the latest version within the constraint, or just ensure the constraint is satisfied? I don't have a strong preference here, but it seems that we are trending toward minimal overrides of user choices and it seems to me that not upgrading follows that most closely - we express a constraint and if that's satisfied, there's nothing to do. If we need an upgrade, that should be reflected in the constraint. I think it's also a perfectly valid choice to say that re-running bootstrap ensures jupyterhub is up-to-date within the constraint. I don't think I'm helping here! We just have to make a choice, and both seem quite sensible.

I don't really see how either choice affects maintenance work. Is it the fact that the `==4.*` constraint plus assumed `--upgrade` gives us an _implicit_ lower bound on the latest `4.*` at any given point in time? So while we express the _install_ constraint, we can in practice have dependencies on newer lower bounds without specifying that constraint anywhere?

re: jupyterhub pinning style - there's no difference whatsoever between `==4.*` and `>=4.0,<5`. They resolve identically. We have to use `>=` if/when we want a minimum minor version, but I don't think how exactly we spell equivalent constraints matters.

manics: Could we make the versions of JupyterHub in the hub and user environments match exactly, i.e. `pip install jupyterhub==$(/opt/tljh/hub/bin/jupyterhub --version)`?

If upgrading JupyterHub to a minor/patch release in the user environment is likely to be disruptive because JupyterHub has increased a minimal version of a dependency then I think that's something we should consider in the JupyterHub repo before making that change. If it's disruptive for TLJH users it's disruptive for other deployments too.

Getting off-topic, but this is one of the reasons I lean towards major version bumps when making changes in JupyterHub that aren't breaking from a sem-ver perspective e.g. dropping Python versions, database upgrades, bumping major dependencies. It helps communicate that "this is a potentially disruptive change", and extends semver to encompass users and administrators rather than just developers. On the other hand it's not that simple because if everything is a major release then it becomes meaningless.

consideRatio: Hmmmm, btw is it in scope to manage manage `pip` version in the user env as well? Otherwise, we may need to think a bit on the implications of `pip install jupyterhub` there as well and its behavior.

consideRatio:
From live meeting chat with input from Min and Simon
- We should `pip install pip>=(current latest)` in both hub env and user env.

tljh/installer.py
consideRatio: I tested pinning to the exact version in the hub env, but we had tests of ensure_user_env that failed, and we installed the user env before the hub env - making it fail due to that as well.

So, instead I opted for reading a new requirements-hub-env.txt file.

consideRatio:
@minrk @manics modified the git history of this PR to make it a bit more slimmed and easy to follow after trying some things and going back on them.

I've updated the title and PR description to reflect this.

Note that it was messy to inspect the hub env as we install the hub env after user env, and that we have unit test for `ensure_user_environment` function that runs without assumptions on a hub env. So, I went for `pip install --upgrade <jupyterhub version constraint from hub env requirements file>` instead.

PR #895, opened 2023-05-16
ensure hub env is on $PATH in jupyterhub service

minrk:
required for `--upgrade-db` to work, since it assumes `alembic` is on $PATH (it probably shouldn't, but that's a PR for jupyterhub)

### DIFF
--- a/tljh/systemd-units/jupyterhub.service
+++ b/tljh/systemd-units/jupyterhub.service
@@ -15,6 +15,7 @@ PrivateDevices=yes
 ProtectKernelTunables=yes
 ProtectKernelModules=yes
 Environment=TLJH_INSTALL_PREFIX={install_prefix}
+Environment=PATH={install_prefix}/hub/bin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 # Run upgrade-db before starting, in case Hub version has changed
 # This is a no-op when no db exists or no upgrades are needed
 ExecStart={python_interpreter_path} -m jupyterhub.app -f {jupyterhub_config_path} --upgrade-db


minrk:
I would think this would cause upgrade from 0.2.0 to fail (because the upgrade does in fact fail in practice). I'm not sure why the upgrade _test_ doesn't fail more reliably. Maybe this is the cause of the intermittent failure [here](https://github.com/jupyterhub/the-littlest-jupyterhub/actions/runs/4992149638/jobs/8939560712#step:5:48)?

```
Downloading traefik 2.10.1 from https://github.com/traefik/traefik/releases/download/v2.10.1/traefik_v2.10.1_linux_amd64.tar.gz...
System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to connect to bus: Host is down
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/srv/src/tljh/installer.py", line 525, in <module>
    main()
  File "/srv/src/tljh/installer.py", line 514, in main
    ensure_jupyterhub_service(HUB_ENV_PREFIX)
  File "/srv/src/tljh/installer.py", line 65, in ensure_jupyterhub_service
    systemd.reload_daemon()
  File "/srv/src/tljh/systemd.py", line 16, in reload_daemon
    subprocess.run(["systemctl", "daemon-reload"], check=True)
  File "/usr/lib/python3.10/subprocess.py", line 524, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['systemctl', 'daemon-reload']' returned non-zero exit status 1.
FAILED
```

PR #906, opened 2023-05-29
Google Cloud troubleshooting and configuration updates

jrdnbradford:
This PR:
- adds a new page on Google Cloud common configuration tasks that details increasing your boot disk size
- adds details on boot disk issues and how to access VM serial port logs to the Google Cloud troubleshooting page

I hope to continue filling out these Google Cloud configuration and troubleshooting pages.


consideRatio:
@jrdnbradford wiee nice!!

My main consern when reviewing docs is that its plausible to maintain long term and won't cause trouble by existing if its not getting maintained. Looking at this, I'm not so concerned because its quite clearly scoped under clear titles and sections! Nice!!!

Skimming through, I think perhaps it could be good to lead with some introduction sentences, for example the how to on increasing the VM's boot disk size could clarify with an intro sentence that the boot disk is a disk associated with the VM where the user storage also resides, and therefore one may want to increase it to provide users with more storage space.

---

If someone wants to help contribute with review effort, please do! Let's try get this merged next week no matter what.

consideRatio:
@jrdnbradford I'm also super happy to see that the images you included were quite small in kB, making the git repo easier to fetch long term.

Thank you!!!

PR #910, opened 2023-06-01
Update systemdspawner from version 0.17.* to 1.*

### LINKED ISSUE #904: Bump to systemdspawner 1.0.0 (not yet released), investigate how to handle sudo disabled by default
In https://github.com/jupyterhub/systemdspawner/pull/91 `SystemdSpawner.disable_user_sudo` was changed to default to `True`, which may be a breaking change we must consider.

### Action points

- [x] Understand the current sudo permissions in tljh
- [x] Understand what sudo permissions we want for tljh 1.0.0 by default
- [x] Take action

consideRatio:
The breaking changes in SystemdSpawner are listed as:

>- Systemd v243+ is now required, and v245+ is recommended. Systemd v245 is available in for example Ubuntu 20.04+, Debian 11+, and Rocky/CentOS 9+.
>- Python 3.8+, JupyterHub 2.3.0+, and Tornado 5.1+ is now required.
>- `SystemdSpawner.disable_user_sudo` (influences systemd's `NoNewPrivileges`) now defaults to `True`, making the installation more secure by default.

I think only the `disable_user_sudo=True` as default change is the only possibly breaking change. I've added a commit to make admin users get it set to False and non-admin users get it set to True explicitly in TLJH now. This leads to the following breaking change:

If a non-admin jupyterhub user has been granted sudo permissions, without being granted jupyterhub admin status, then the user would then not be able to sudo any more.

I suggest that we suggest in the changelog that such users should also be granted jupyterhub admin status. This is something the user could acquire with sudo anyhow.

- Closes #904

### DIFF
--- a/tljh/requirements-hub-env.txt
+++ b/tljh/requirements-hub-env.txt
@@ -9,7 +9,7 @@
 # version release of tljh.
 #
 jupyterhub>=4.0.0,<5
-jupyterhub-systemdspawner>=0.17.0,<5
+jupyterhub-systemdspawner>=1.0.0,<2
 jupyterhub-firstuseauthenticator>=1.0.0,<2
 jupyterhub-nativeauthenticator>=1.2.0,<2
 jupyterhub-ldapauthenticator>=1.3.2,<2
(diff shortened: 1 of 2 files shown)


consideRatio:
- Test failures on main branch as well, see #913

PR #912, opened 2023-06-02
Stop bundling jupyterhub-configurator which has been disabled by default

### LINKED ISSUE #901: Remove jupyterhub-configurator in 1.0.0?
I'd like to propose we remove the opt-in setup of jupyterhub-configurator.

### Overview

According to the docs, the configurator is disabled by default but is documented as something that can be [experimetnally enabled](https://github.com/jupyterhub/the-littlest-jupyterhub/blob/HEAD/docs/topic/jupyterhub-configurator.md#L1-L21).

If it is enabled, the Spawner get's an imported ConfiguratorSpawnerMixin that checks the relevant configuration just in time.

https://github.com/jupyterhub/the-littlest-jupyterhub/blob/c6c3c96f9aa21a8dcfee594db0c0f9f180193f40/tljh/user_creating_spawner.py#L41-L50

What can be configured by the jupyterhub configurator is hardcoded in a tljh package bundled config file.

https://github.com/jupyterhub/the-littlest-jupyterhub/blob/c6c3c96f9aa21a8dcfee594db0c0f9f180193f40/tljh/configurer.py#L278-L304

That declares that the default url can be configured:

https://github.com/jupyterhub/the-littlest-jupyterhub/blob/c6c3c96f9aa21a8dcfee594db0c0f9f180193f40/tljh/jupyterhub_configurator_config.py#L1

### Proposal motivations

- We are heading for a 1.0.0 release and can do breaking changes as part of it
- Enabling jh-configurator via tljh is hardcoded to only allow config of the default interface (`default_url` ?)
- As a tljh jupyterhub admin one can use either the configurator to set default_url specifically, or use `tljh-config` or plain jupyterhub config `.py` files to do ~anything without the configurator. Should we really teach jh admins another way of doing things just to configure default-url in a simpler way?
- jupyterhub-configurator is not versioned, so we reference fixed git hashes currently
- `tljh-config` is already a shortcut to configure things by adding another configuration layer, adding another with the configurator can make it harder to understand how things work overall

consideRatio:
This removal PR was quite easy to do so I went for it, hopefully it helps evaluate if it makes sense to remove as proposed in #901.

- Closes #901


consideRatio:
- Test failures on main branch as well, see #913

yuvipanda (commented):
I'd very much like @GeorgianaElena's opinion, as she worked on integrating it with TLJH

consideRatio:
Thank you @GeorgianaElena @yuvipanda @manics for chiming in about this! :heart: :tada:

I'll go for a merge if the tests pass picking up a consensus about removing it for now.

PR #916, opened 2023-06-06
maint: refactor tests, fix upgrade tests (now correctly failing)

### LINKED ISSUE #899: Test logs from tests extremely hard to read
Typically I expect running tests to just show what test is being run, and then when the tests have been run, to see captured output associated with a failure if there is a failure.

With our test setup though, we get output streamed and everything becomes garbled. For example I know I'm not the only one to have been confused about seeing an error about not starting as PID 1 - but that was just a test that verified that we error nicely when starting as something else than PID 1 etc.

Here is another example where two tests pass run and we see the logs from that streamed. The first test is running something expected to fail - which it does (XFAIL), and one that is expected to succeed - which it does (PASSED). Skimming through this, one would think something is wrong - but its not, so its just a distraction.

image

I expected this to look more like this:

image

### Suggested action

- [ ] Let's capture relevant logs while a test is running, and then emit it on failure like typically done.
- [ ] Let's configure pytest with `--max-fail=1` to quicker see information about a failed tests, and make debugging it easier. We won't see all failures, but we may improve our ability to understand the first failure we encounter and get the captured logs presented faster.

consideRatio:
I have various soft goals I'm working towards

1. Make the tests and the infrastructure to run them a bit easier to understand and maintain
2. Make run tests emit easier to understand results if they succeed or fail
   Closes #899
3. Make tests runs a bit faster to complete if possible
4. Capture more test coverage if possible

I'm still deliberating on copying files to the container vs mounting the local git repo in the container. I figure maybe we can get test coverage by mounting instead of copying, but its also coupling things a bit.

### Important insights

1. Our upgrade didn't work before, now they do.
   We have a real test failure though, and I don't aim to solve it as part of this PR. It relates to conda/mamba and tsting the `tljh_extra_user_conda_packages` plugin, which indicates there is an issue in the conda/mamba state after the upgrade to a new conda/mamba version.

### Misc things not done

- [ ] Let bootstrap.py logs and tljh.installer logs clarify when the bootstrap.py script stops running and the tljh.installer starts running
- [ ] Stop copying files to integration test container, and mount things instead to get test coverage?
- [ ] Fix test failure discovered by fixed upgrade test
- [ ] Let our test suite run with JupyterHub.log_level set to debug


consideRatio:
I've invested a lot of time into this PR and I could keep going because there are more things to do, but I'd love to get this merged and go from there with smaller increments going onwards.

The two test failures are real, but not caused by this PR besides fixing the tests to not be incorrectly passing because they failed to start from an older install. I've opened https://github.com/jupyterhub/the-littlest-jupyterhub/issues/918 to represent this and hope that we can get this merged without resolving that as part of this PR.

minrk:
This is wonderful! Happy to merge and move on.

consideRatio:
Thank you @minrk!!!!!

PR #920, opened 2023-06-09
--force-reinstall old conda to ensure it's working before we try to install conda packages

### LINKED ISSUE #918: Upgrade from 0.2.0 fails (because a corrupt conda/mamba file?)
In #916 I fixed the upgrade tests that wasn't upgrading from the 0.2.0 version as they should, but that uncovered [a test failure when upgrading from 0.2.0 (aka. latest)](https://github.com/jupyterhub/the-littlest-jupyterhub/actions/runs/5217412600/jobs/9417219661#step:7:62).

I suspect its related to this warning.

```
WARNING Extracted package cache '/opt/tljh/user/pkgs/libxml2-2.9.12-h72842e0_0' has invalid 'repodata_record.json' file and will be removed: [json.exception.type_error.302] type must be string, but is null
```

Even though the actual error comes later, and is:

```
RemoveError: 'requests' is a dependency of conda and cannot be removed from conda's operating environment.
```

minrk:
an old conda bug causes RemoveError: requests is a dependency of conda when installing other packages (closes #918).

<del>
Upgrading conda itself first avoids this bug (we _may_ want `--force` here, which some folks have seen required when conda itself is messed up, but let's avoid it for now). So we know that the version in 0.2.0 (4.10) is not actually new enough to work. It's unclear what the true minimum version is to fix this, but the important thing is that it will be upgraded if it's still the version in 0.2.0 (4.10), so pick the versions from today which we know works, and will be the version upgraded-to if an 0.2 install has yet to update conda.
</del>

`conda install --force-reinstall conda` is the fix for this, because `conda install conda` will fail to upgrade if we're in this situation. We could make it conditional on old condas, as I don't think this is likely to come up with more recent versions, but it should be safe in general to run this command.

### DIFF
--- a/tljh/installer.py
+++ b/tljh/installer.py
@@ -148,8 +148,8 @@ MAMBAFORGE_CHECKSUMS = {
 # minimum versions of packages
 MINIMUM_VERSIONS = {
     # if conda/mamba/pip are lower than this, upgrade them before installing the user packages
-    "mamba": "0.16.0",
-    "conda": "4.10",
+    "mamba": "1.4.2",
+    "conda": "23.3.1",
     "pip": "23.1.2",
     # minimum Python version (if not matched, abort to avoid big disruptive updates)
     "python": "3.9",
--- a/tljh/installer.py
+++ b/tljh/installer.py
@@ -148,8 +148,8 @@ MAMBAFORGE_CHECKSUMS = {
 # minimum versions of packages
 MINIMUM_VERSIONS = {
     # if conda/mamba/pip are lower than this, upgrade them before installing the user packages
-    "mamba": "1.4.2",
-    "conda": "23.3.1",
+    "mamba": "0.16.0",
+    "conda": "4.10",
     "pip": "23.1.2",
     # minimum Python version (if not matched, abort to avoid big disruptive updates)
     "python": "3.9",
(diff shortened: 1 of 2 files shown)


minrk:
If we don't want to bump versions, it appears the fix is specifically:

```
conda install --force-reinstall conda
```

which should ensure that conda itself is in a consistent state (and should not upgrade it). We could do that as a _separate_, unconditional 'ensure conda is okay' step before installing anything, but I believe using `--force-reinstall` to ensure that conda/mamba are super consistent ought to be good enough.

We could also try to restrict 'needs force' to a version constraint, but I have no idea what that constraint would be.

consideRatio:
> We could also try to restrict 'needs force' to a version constraint, but I have no idea what that constraint would be.

I think its great if we avoid introducing that precautionary, as it makes the situations that can arise grow and its harder to ensure function in all situations. For the same reasons, I'm happy about seeing more stringent requirements on the conda/mamba versions.

minrk:
`--force-reinstall` behaves a bit weird. It doens't upgrade packages, so unconstrained versions result in no upgrade unless the version constraint is in the install command.

I think maybe it makes sense to _unconditionally_ force-reinstall conda/mamba to ensure it's always consistent, followed by the upgrade steps.

minrk:
Struggling to get the right behavior here to match elaborate test expectations. It seems `--force-reinstall` _does_ upgrade when run with old versions, unlike more recent ones.

It really seems the 'reinstall if needed' is what we want, but I can't seem to figure out 'if needed'.

tljh/installer.py
```diff
@@ -242,21 +242,42 @@ def ensure_user_environment(user_requirements_txt_file):
...
-            conda.ensure_pip_packages(
-                USER_ENV_PREFIX,
-                pypi_pkgs_to_upgrade,
-                upgrade=True,
-            )
+    # force reinstall conda/mamba to ensure a basically consistent env
+    # avoids issues with RemoveError: 'requests' is a dependency of conda
+    # only do this for 'old' conda versions known to have a problem
+    # we don't know how old, but we know 4.10 is affected and 23.1 is not
+    if not is_fresh_install and V(package_versions.get("conda", "0")) < V("23.1"):
```
minrk, line 249: I don't really like what I ended up doing here, but I bracketed the force reinstall on a version _older_ than what's going to be in 1.0, with the hope that we actually remove this step when we drop support for upgrading from earlier than 1.0.

I think you need pretty old conda to get here, but that's what we have in 0.2. It really is a special-case to handle:

- quite old conda
- inconsistent environment in conda's own dependencies (caused by our own `pip install --upgrade` commands)

but this is what you'll get if you start with tljh 0.2 and haven't upgraded conda, so we _should_ handle it. I think it is appropriate to treat it specially, as if users have broken their own envs such that conda itself doesn't work, I think it's appropriate for us to not handle that and require users to fix their conda (or delete their user env to start fresh) before upgrading. But since _we_ are creating a broken environment, we need to handle it, at least for the very specific case of upgrading from a previous tljh known to have this issue.

consideRatio (approved):
LGTM, go for merge?

minrk:
_phew_ everything's passing and test expectations are unchanged.

One thing that might help these things (or might not, honestly, since it creates an extra env to manage) is to make the root conda install a separate environment from the user env:

- conda root env - /opt/tljh/conda
- user env created with `/opt/tljh/conda/bin/conda create --prefix /opt/tljh/user`

Then conda/mamba themselves won't actually interact with user packages. This is where the `pip install` step caused problems by upgrading requests, a dependency of conda itself. There would be no reason to have pip packages in the root env if we did this.

minrk:
all set from my side, yes

yuvipanda:
btw, I'm still seeing this on https://github.com/jupyterhub/the-littlest-jupyterhub/actions/runs/6357501669/job/17268703149#step:5:157 from https://github.com/jupyterhub/the-littlest-jupyterhub/pull/942 (which is based on latest main)

PR #928, opened 2023-08-01
Update Notebook, JupyterLab, Jupyter Resource Usage

jtpio:
Update Jupyter Notebook, JupyterLab and Jupyter Resource Usage to the latest major releases:

- [x] `notebook`: https://jupyter-notebook.readthedocs.io/en/stable/changelog.html
- [x] `jupyterlab`: https://jupyterlab.readthedocs.io/en/latest/getting_started/changelog.html#v4-0
- [x] `jupyter-resource-usage`: https://github.com/jupyter-server/jupyter-resource-usage/releases/tag/v1.0.0
- [x] Pending a new `nbgitpuller` release: https://github.com/jupyterhub/nbgitpuller/pull/315

### DIFF
--- a/integration-tests/test_extensions.py
+++ b/integration-tests/test_extensions.py
@@ -1,3 +1,4 @@
+import re
 import subprocess
 
 
--- a/integration-tests/test_extensions.py
+++ b/integration-tests/test_extensions.py
@@ -7,7 +7,7 @@ def test_serverextensions():
     """
     # jupyter-serverextension writes to stdout and stderr weirdly
     proc = subprocess.run(
-        ["/opt/tljh/user/bin/jupyter-serverextension", "list", "--sys-prefix"],
+        ["/opt/tljh/user/bin/jupyter-server", "extension", "list", "--sys-prefix"],
         stderr=subprocess.PIPE,
     )
 
--- a/integration-tests/test_extensions.py
+++ b/integration-tests/test_extensions.py
@@ -41,6 +41,6 @@ def test_labextensions():
     for e in extensions:
         # jupyter labextension lists outputs to stderr
         out = proc.stderr.decode()
-        enabled_ok_pattern = re.compile(fr"{e}.*enabled.*OK")
+        enabled_ok_pattern = re.compile(rf"{e}.*enabled.*OK")
         matches = enabled_ok_pattern.search(out)
         assert matches is not None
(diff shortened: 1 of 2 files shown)


consideRatio:
Thank you @jtpio!

This won't be a breaking change as things aren't automatically upgraded, this is just for new installations anyhow. With that in mind I'm thinking this PR should be merged before ASAP and before 1.0.0 release.

consideRatio:
I see integration tests are failing because a call to a binary no longer around. But this test will be run with new and old versions of software installed in the user environment, because during an upgrade test we install with the previous version first and then install again with a new version, where the user environment isn't updated.

So fixing this test failure must be done in a way supporting either binary name I think.

jtpio:
> This won't be a breaking change as things aren't automatically upgraded,

Ah that's good news!

Also this change might need a new release of `nbgitpuller` with https://github.com/jupyterhub/nbgitpuller/pull/240

consideRatio:
@jtpio we could relax this test I think, and just go for `jupyterlab` and `nbgitpuller` instead of `jupyterlab 4.` etc

consideRatio:
> Also this change might need a new release of `nbgitpuller` with [jupyterhub/nbgitpuller#240](https://github.com/jupyterhub/nbgitpuller/pull/240)

I opened https://github.com/jupyterhub/nbgitpuller/pull/315 about this

jtpio:
> I opened [jupyterhub/nbgitpuller#315](https://github.com/jupyterhub/nbgitpuller/pull/315) about this

Nice thanks!



> So fixing this test failure must be done in a way supporting either binary name I think, or skipping it if the new binary isn't available - that is also fine in my mind.

Checking first if the tests pass when using the new `jupyter server extension` and `jupyter labextension list` commands. If we don't install `notebook==6.*` or `nbclassic==1.*` in the environment then the old `jupyter serverextension list` should not be available anymore.

Actually even with `nbclassic` the command would now be `jupyter-nbclassic-serverextension`: https://nbclassic.readthedocs.io/en/latest/nbclassic_dev_faq.html#noteworthy-updates-in-nbclassic

jtpio:
Question: should users be able to access the classic notebook UI? If yes:

- we could add `nbclassic==1.*` to the user environment. The classic notebook UI would be available under `/nbclassic`
- we could document how to install `nbclassic` in the user environment, for example in https://tljh.jupyter.org/en/latest/howto/user-env/user-environment.html.

consideRatio (approved):

consideRatio:
Since this is for new installs etc, I suggest we make it not come with more stuff.

I'm prefer seeing tljh install as little as possible though to let the admins decide from a relatively clean slate if they install more things etc. I think for now, lets go with this.

PR #935, opened 2023-08-12
adapt install documentation for new /lab default interface

schwebke:
Adapt fresh install documentation for new default `/lab` interface since 1.0.0.


minrk:
thanks!

PR #966, opened 2024-02-24
ci: add test for debian 12

consideRatio:
Ended up doing many things, and that led to spotting things needs doing in hubtraf and idle-culler. Draft until i can refine this


consideRatio:
Broken apart to
- #967
- #968 
- maybe more..

PR #968, opened 2024-02-26
Re-install conda/mamba for every tljh upgrade (doesn't imply upgrade)

### LINKED ISSUE #944: Test suite is failing in main branch
I've not looked into this further =/

consideRatio:
We have previously re-installed (without upgrading) conda and mamba, and done it based on specific versions. I think we are required to do this uncodntionally though. I think based https://github.com/conda/conda/issues/12366#issuecomment-1466144007 that since tljh make use of both pip and conda/mamba to install packages via plugins, even though we mostly use pip, we must do this re-install to ensure conda/mamba is functional before an tljh upgrade.

I think we end up with `requests` installed via pip during the initial setup of tljh, and that sets us up for issues when we are to make an upgrade later when conda/mamba is being used via plugins. I think the reason we didn't have this error before in tests, but do now, relates to new versions of `openssl` have been released, and that conda/mamba automatically upgrades them.

- Fixes #944

### DIFF
--- a/.github/integration-test.py
+++ b/.github/integration-test.py
@@ -167,7 +167,15 @@ def run_test(
         command = f"python3 /srv/src/bootstrap/bootstrap.py --version={upgrade_from}"
         run_command(container_name, command)
 
+        # show user environment
+        command = "/opt/tljh/user/bin/mamba list"
+        run_command(container_name, command)
+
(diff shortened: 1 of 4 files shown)


consideRatio:
@manics @minrk this may be relevant to ensure upgrade works, especially soon if users are to use the upgrade to get security patches or similar.

PR #969, opened 2024-03-01
Fix typo and replace word

davidalber:
This PR fixes a typo and replaces the word. The typo is of "distinguishing", but "distinctive" makes more sense to me in the context of the sentence it is in. Let me know if you prefer just fixing the typo and leaving "distinguishing".

Before (from [Installing on Amazon Web Services](https://tljh.jupyter.org/en/latest/install/amazon.html)):

After:


consideRatio:
Thank you @davidalber!!

PR #975, opened 2024-04-03
tests: fix to catch test failure earlier when they really happen

### LINKED ISSUE #974: Test failures in main branch
Debugged in #975

Partially fixed in #962 by @jrdnbradford!

consideRatio:
Just something observed when trying to resolve #974, which is fixed by jupyterhub 4.1.5 release and work in #962 by @jrdnbradford.

### DIFF
--- a/integration-tests/test_hub.py
+++ b/integration-tests/test_hub.py
@@ -59,9 +59,9 @@ async def test_user_code_execute():
 
     async with User(username, HUB_URL, partial(login_dummy, password="")) as u:
         assert await u.login()
-        await u.ensure_server_simulate(timeout=60, spawn_refresh_time=5)
-        await u.start_kernel()
-        await u.assert_code_output("5 * 4", "20", 5, 5)
+        assert await u.ensure_server_simulate(timeout=60, spawn_refresh_time=5)
+        assert await u.start_kernel()
+        assert await u.assert_code_output("5 * 4", "20", 5, 5)
 
 
 async def test_user_server_started_with_custom_base_url():


integration-tests/test_hub.py
```diff
@@ -59,9 +59,9 @@ async def test_user_code_execute():
 
     async with User(username, HUB_URL, partial(login_dummy, password="")) as u:
         assert await u.login()
-        await u.ensure_server_simulate(timeout=60, spawn_refresh_time=5)
-        await u.start_kernel()
-        await u.assert_code_output("5 * 4", "20", 5, 5)
+        assert await u.ensure_server_simulate(timeout=60, spawn_refresh_time=5)
+        assert await u.start_kernel()
+        assert await u.assert_code_output("5 * 4", "20", 5, 5)
```
consideRatio, line 64: Can return False when failing, so we should error early if that happens.

consideRatio:
@minrk I've concluded that when we upgrade from 0.2.0, the user environment will be old in general, and our tests with jupyterhub 4.1+ fails when hubtraf [tries to start a kernel](https://github.com/yuvipanda/hubtraf/blob/2be1d3558b2f0925d4dfb169cdc4b89b010c8952/hubtraf/user.py#L263-L265) with a request to `/user/someuser/api/kernels` and passing headers that include `X-XSRFToken` thanks to [earlier starting the server](https://github.com/yuvipanda/hubtraf/blob/2be1d3558b2f0925d4dfb169cdc4b89b010c8952/hubtraf/user.py#L227-L229).

The version in the user env for server components are `notebook==6.5.6` and `jupyter_server==1.24.0`, which are the latest notebook 6 and jupyter_server 1 versions - is this failure expected?

minrk:
Not expected, but since hubtraf is emulating a browser, it might not be doing everything browsers do (in particular, the valid XSRF token _changes_ over time, and it may be caching an old value). Does it only happen with the oldest versions, not newer ones?

Having a look now.

minrk:
Found it: https://github.com/jupyterhub/jupyterhub/pull/4771

I can make 4.1.5

consideRatio:
Wieeeee nice @minrk!!

minrk:
testing here with the 4.x branch, then I'll release 4.1.5: https://github.com/jupyterhub/jupyterhub/pull/4773

consideRatio:
Unrelated other failure - success i think!

consideRatio:
The previously broken 0.2.0 upgrade tests is no longer broken - thank you @minrk!!

PR #976, opened 2024-04-06
Add TLJH config lockfile

jrdnbradford:
This adds a lockfile for the TLJH config. Given the reused code it seems things are due for a refactor. I partially did this by abstracting the code that retrieves the current config.

### DIFF
--- a/tljh/requirements-hub-env.txt
+++ b/tljh/requirements-hub-env.txt
@@ -27,4 +27,5 @@ jupyterhub-idle-culler>=1.2.1,<2
 #
 pycurl>=7.45.2,<8
 
-filelock>=3.13.3,<4
+# filelock is used to help us do atomic operations on config file(s)
+filelock>=3.15.4,<4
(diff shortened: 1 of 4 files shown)


tljh/requirements-hub-env.txt
```diff
@@ -26,3 +26,5 @@ jupyterhub-idle-culler>=1.2.1,<2
 # ref: https://github.com/jupyterhub/the-littlest-jupyterhub/issues/289
 #
 pycurl>=7.45.2,<8
+
+filelock>=3.13.3,<4
```
consideRatio, line 30:
```suggestion
# filelock is used to help us do atomic operations on config file(s)
filelock>=3.15.4,<4
```

consideRatio (approved):
Thank you @jrdnbradford!! I see that this resolves a FIXME comment on line 212.

I considered perhaps lifting the import `from filelock import FileLock, Timeout` from the function scope to the top, but I'm not sure if there was a reason about having it within the function. I figure its not critical though, so we could merge anyhow I think.

@minrk what do you think about this, OK to merge?

jrdnbradford:
Awesome @consideRatio. It's been a while, but if I recall correctly `from filelock import FileLock, Timeout` caused issues outside the function scope. I _think_ it either caused tests to fail and/or an import error. Mostly I just remember being a bit flummoxed that I couldn't get it to work outside the function scope. 🥈

consideRatio:
Thanks for following up, and thank you again for working this!

PR #983, opened 2024-05-23
Added missing details on how to add custom domain from manual HTTPS  configuration

josedaudi:
While trying to configure manual HTTPS for TLJH i found that details on how to had custom domain was missing, for easy understanding and clarity had to propose this change.


yuvipanda:
Apologies for the late response, and thank you for taking the time to contribute here, @josedaudi!

PR #989, opened 2024-08-23
jupyterhub 5

minrk:
require 5.1.0 for security fixes


minrk:
FirstUse should have allow_all = True by default, I think: https://github.com/jupyterhub/firstuseauthenticator/pull/56

manics (approved):

PR #990, opened 2024-08-23
Update base user environment to miniforge 24.7.1-0 (Python 3.12)

minrk:
mambaforge-named installers are deprecated, as mamba has been in miniforge for some time now, so they will be merged. Mambaforge releases will stop later this year.


consideRatio:
Wieee super thankful this is done @minrk! I was thinking we should do something like this before tljh v2 and you have already done it! :D

I pushed a commit bumping to miniforge of a version released yesterday

PR #994, opened 2024-09-05
consolidate lock file handling

minrk:
- makes filelock a dependency of tljh (missing this is the reason for the in-function imports)
- adds config_file_lock context manager to re-use
- moves filelock import to module level
- fixes exit codes, stderr messages

small polish on #976

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,4 +1,3 @@
-filelock
 packaging
 pytest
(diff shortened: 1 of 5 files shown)


jrdnbradford:
So that's what I missed in #976. I kept throwing `filelock` into requirements files when it should have gone into `setup.py`. Thanks @minrk.

PR #998, opened 2024-09-20
Drop ubuntu 20.04, require py39, traefik-proxy v2, and ldapauthenticator v2

### LINKED ISSUE #996: jupyterhub/traefik-proxy v2 is out, we should bump to use it!
It seems that the only breaking changes was requiring python 3.8 - see https://jupyterhub-traefik-proxy.readthedocs.io/en/latest/changelog.html.

consideRatio:
- fixes #996

I propose we drop support for ubuntu 20.04, which is more than 4 years old now, and python 3.8 having reached end of life. With tljh v2 coming up, it seems like a good time to do it.


minrk (approved):

consideRatio:
Updated to ldapauthenticator 2.0.0b2 from 2.0.0b1

Issue #991, opened 2024-09-04
Release planning for v2


consideRatio:
I'd like to work towards a v2 release with jupyterhub 5 and oauthenticator 17.

## Todo

- #990
- Test and verify function against ubuntu 24.04
  - [x] Integration tests
  - [x] Unit tests: https://github.com/jupyterhub/the-littlest-jupyterhub/pull/993
- [x] Transition to jupyterhub 5 (#989)
- [x] Transition to oauthenticator 17
  - [x] Note that `oauthenticator[azuread]>=16.0.4,<17` is installed, but we no longer need to specify `azuread` I think. Look  it up in oauthenticator.
  - #992 
- Consider getting a release out for:
  - [x] [jupyterhub-systemdspawner](https://github.com/jupyterhub/systemdspawner)>=1.0.1,<2
    - [x] Ensure tests for jupyterhub 5 etc pass and refresh CI system things
      https://github.com/jupyterhub/systemdspawner/pull/143
    - [x] No merged PRs, not aiming for a release.
  - [x] [jupyterhub-firstuseauthenticator](https://github.com/jupyterhub/firstuseauthenticator)>=1.0.0,<2
    - [x] Ensure tests for jupyterhub 5 etc pass and refresh CI system things
      https://github.com/jupyterhub/firstuseauthenticator/pull/57
      https://github.com/jupyterhub/firstuseauthenticator/pull/56
    - [x] Make release, a PR related to jupyterhub 5 is merged.
      https://github.com/jupyterhub/firstuseauthenticator/pull/58
  - [x] [jupyterhub-nativeauthenticator](https://github.com/jupyterhub/nativeauthenticator)>=1.2.0,<2
    - [x] Ensure tests for jupyterhub 5 etc pass and refresh CI system things
      https://github.com/jupyterhub/nativeauthenticator/pull/274
    - Fix a bug: https://github.com/jupyterhub/nativeauthenticator/pull/275
    - [x] Make release, things have happened since last.
      
  - [x] [jupyterhub-ldapauthenticator](https://github.com/jupyterhub/ldapauthenticator)>=1.3.2,<2
    - [x] Ensure tests for jupyterhub 5 etc pass and refresh CI system things
      https://github.com/jupyterhub/ldapauthenticator/pull/244
      https://github.com/jupyterhub/ldapauthenticator/pull/245
      https://github.com/jupyterhub/ldapauthenticator/issues/246
    - [x] Make release of v2.0.0
[...]

PR #1000, opened 2024-09-26
Add changelog for 2.0.0b1

consideRatio:
- for #991

---

Is there something else to include before going for a 2.0.0b1 release?

---

I've recently self-merged these and included them in this changelog.
- #1001
- #1002

---

Link to rendered changelog: https://the-littlest-jupyterhub--1000.org.readthedocs.build/en/1000/reference/changelog.html

I've not done work including v1 -> v2 upgrade instructions. My take is that its relevant to do when there is a distribution specific migration that needs to be done for example, but I think that isn't the case for this major version of tljh, so I've leaned on referencing other software's changelogs.

### DIFF
--- a/docs/reference/changelog.md
+++ b/docs/reference/changelog.md
@@ -2,6 +2,129 @@
 
 # Changelog
 
+## 2.0
+
+### 2.0.0b1 - 2024-09-30
+
+This release bundles with the latest available software from the JupyterHub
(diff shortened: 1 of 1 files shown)


minrk (approved):

minrk:
Thanks @consideRatio !

PR #1003, opened 2024-09-28
Test upgrade from 1.*

manics:
Currently covered by `latest`, but won't be when 2.0.0 is released


consideRatio:
I missed adding this into the changelog, should be done later. Making a note in the release planning issue.

PR #1013, opened 2024-11-14
Fix `cpu` and `memory` config validation in JSON schema

### LINKED ISSUE #1012: tljh-config set limits.cpu does not accept a float value
### Bug description

Trying to set the `limits.cpu` configuration to a `float` raises a validation error.

### How to reproduce

`sudo tljh-config set limits.cpu 3.5`

#### Expected behaviour

According to the documentation: <https://tljh.jupyter.org/en/latest/topic/tljh-config.html#tljh-set-user-limits> a float is a valid value.

### Your personal set up

 - OS: Ubuntu 24.04.1 LTS
 
 - Version(s): (The Littlest) Jupyterhub: 5.2.1, Python: 3.12.3
 

<details><summary>Full environment</summary>

```
attrs==23.2.0
Automat==22.10.0
Babel==2.10.3
bcc==0.29.1
bcrypt==3.2.2
blinker==1.7.0
certifi==2023.11.17
chardet==5.2.0
click==8.1.6
cloud-init==24.3.1
colorama==0.4.6
command-not-found==0.3
configobj==5.0.8
constantly==23.10.4
cryptography==41.0.7
dbus-python==1.3.2
distro==1.9.0
distro-info==1.7+build1
httplib2==0.20.4
hyperlink==21.0.0
idna==3.6
incremental==22.10.0
Jinja2==3.1.2
jsonpatch==1.32
jsonpointer==2.0
jsonschema==4.10.3
launchpadlib==1.11.0
lazr.restfulclient==0.14.6
lazr.uri==1.0.6
markdown-it-py==3.0.0
MarkupSafe==2.1.5
mdurl==0.1.2
netaddr==0.8.0
netifaces==0.11.0
oauthlib==3.2.2
pexpect==4.9.0
ptyprocess==0.7.0
pyasn1==0.4.8
pyasn1-modules==0.2.8
Pygments==2.17.2
PyGObject==3.48.2
PyHamcrest==2.1.0
PyJWT==2.7.0
pyOpenSSL==23.2.0
pyparsing==3.1.1
pyrsistent==0.20.0
pyserial==3.5
python-apt==2.7.7+ubuntu3
python-debian==0.1.49+ubuntu2
python-magic==0.4.27
pytz==2024.1
PyYAML==6.0.1
requests==2.31.0
rich==13.7.1
service-identity==24.1.0
setuptools==68.1.2
six==1.16.0
sos==4.5.6
ssh-import-id==5.11
systemd-python==235
Twisted==24.3.0
typing_extensions==4.10.0
ubuntu-pro-client==8001
ufw==0.36.2
unattended-upgrades==0.1
urllib3==2.0.7
wadllib==1.3.6
wheel==0.42.0
zope.interface==6.1
```
</details>

<details><summary>Configuration</summary>

```python
# jupyterhub_config.py
"""
JupyterHub config for the littlest jupyterhub.
"""
import os
from glob import glob
from tljh import configurer
from tljh.config import CONFIG_DIR, INSTALL_PREFIX, USER_ENV_PREFIX
from tljh.user_creating_spawner import UserCreatingSpawner
from tljh.utils import get_plugin_manager
c = get_config()  # noqa
c.JupyterHub.spawner_class = UserCreatingSpawner
c.JupyterHub.cleanup_servers = False
c.JupyterHub.hub_port = 15001
c.TraefikProxy.should_start = False
dynamic_conf_file_path = os.path.join(INSTALL_PREFIX, "state", "rules", "rules.toml")
c.TraefikFileProviderProxy.dynamic_config_file = dynamic_conf_file_path
c.JupyterHub.proxy_class = "traefik_file"
c.SystemdSpawner.extra_paths = [os.path.join(USER_ENV_PREFIX, "bin")]
c.SystemdSpawner.default_shell = "/bin/bash"
c.SystemdSpawner.unit_name_template = "jupyter-{USERNAME}"
tljh_config = configurer.load_config()
configurer.apply_config(tljh_config, c)
pm = get_plugin_manager()
pm.hook.tljh_custom_jupyterhub_config(c=c)
extra_configs = sorted(glob(os.path.join(CONFIG_DIR, "jupyterhub_config.d", "*.py")))
for ec in extra_configs:
    load_subconfig(ec)

```
</details>

<details><summary>Logs</summary>

```
Config validation error: 3.5 is not of type 'integer'.
You can still apply this change without validation by re-running your command with the --no-validate flag.
If you think this validation error is incorrect, please report it to https://github.com/jupyterhub/the-littlest-jupyterhub/issues.
```
</details>

jrdnbradford:
Fixes #1012. 

The config schema for `cpu` limits now accepts values such as `3`, `3.5` and `"None"`. Added `minimum`s of `0` to prevent negative numerics.

Details on `number` can be found in the [JSON Schema docs](https://json-schema.org/understanding-json-schema/reference/numeric).

### DIFF
--- a/tljh/config_schema.py
+++ b/tljh/config_schema.py
@@ -79,23 +79,16 @@ config_schema = {
             "description": "User CPU and memory limits.",
             "type": "object",
             "additionalProperties": False,
-            "properties": {"memory": {"type": "string"}, "cpu": {
-            "oneOf": [
-                {
-                  "type": "integer",
-                  "minimum": 0
+            "properties": {
+                "memory": {"type": "string"},
+                "cpu": {
+                    "anyOf": [
+                        {"type": "integer", "minimum": 0},
+                        {"type": "number", "minimum": 0},
+                        {"type": "string", "enum": ["None"]},
+                    ]
                 },
-                {
-                  "type": "number",
-                  "minimum": 0
-                },
-                {
-                  "type": "string",
-                  "enum": ["None"]
-                }
-            ]
-          }
-        },
+            },
         },
         "UserEnvironment": {
             "type": "object",
--- a/tljh/config_schema.py
+++ b/tljh/config_schema.py
@@ -83,7 +83,6 @@ config_schema = {
                 "memory": {"type": "string"},
                 "cpu": {
                     "anyOf": [
-                        {"type": "integer", "minimum": 0},
                         {"type": "number", "minimum": 0},
                         {"type": "string", "enum": ["None"]},
                     ]
--- a/tljh/config_schema.py
+++ b/tljh/config_schema.py
@@ -80,7 +80,12 @@ config_schema = {
             "type": "object",
             "additionalProperties": False,
             "properties": {
-                "memory": {"type": "string"},
+                "memory": {
+                    "anyOf": [
+                        {"type": "string"},
+                        {"type": "null"},
+                    ]
+                },
                 "cpu": {
                     "anyOf": [
                         {"type": "number", "minimum": 0},
--- a/tljh/config_schema.py
+++ b/tljh/config_schema.py
@@ -79,7 +79,23 @@ config_schema = {
             "description": "User CPU and memory limits.",
             "type": "object",
             "additionalProperties": False,
-            "properties": {"memory": {"type": "string"}, "cpu": {"type": "integer"}},
+            "properties": {"memory": {"type": "string"}, "cpu": {
+            "oneOf": [
+                {
+                  "type": "integer",
+                  "minimum": 0
+                },
+                {
+                  "type": "number",
+                  "minimum": 0
+                },
+                {
+                  "type": "string",
+                  "enum": ["None"]
+                }
+            ]
+          }
+        },
         },
         "UserEnvironment": {
             "type": "object",
--- a/tljh/config_schema.py
+++ b/tljh/config_schema.py
@@ -84,7 +84,7 @@ config_schema = {
                 "cpu": {
                     "anyOf": [
                         {"type": "number", "minimum": 0},
-                        {"type": "string", "enum": ["None"]},
+                        {"type": "null"},
                     ]
                 },
             },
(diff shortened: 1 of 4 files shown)


tljh/config_schema.py
```diff
@@ -79,7 +79,15 @@
             "description": "User CPU and memory limits.",
             "type": "object",
             "additionalProperties": False,
-            "properties": {"memory": {"type": "string"}, "cpu": {"type": "integer"}},
+            "properties": {
+                "memory": {"type": "string"},
+                "cpu": {
+                    "anyOf": [
+                        {"type": "number", "minimum": 0},
+                        {"type": "string", "enum": ["None"]},
```
minrk, line 87: Thanks!

Shouldn't this one be actual `null`, not the string 'None'?

```suggestion
                        {"type": "null"},
```

jrdnbradford: Ah, looks like this is slightly more complicated than I thought!

> Setting it to `None` disables CPU limits.

I used a string `"None"` based on a quick glance at the above [docs](https://tljh.jupyter.org/en/latest/topic/tljh-config.html#user-server-limits), not realizing that moving between `tljh-config` to Python to YAML and back again might cause issues.

If I use your recommended change, the following commands both fail validation:

```bash
tljh-config set limits.cpu None
tljh-config set limits.cpu null
```

Seems as if a change to `config.py` will also need to be made to convert the string `"None"`, `"null"`, or some other placeholder-type string value passed to `tljh-config` to Python's `None` for validation.

Since most users use `tljh-config` and the docs already reference `None`, maybe `None` is the right value for this, ie users will run 
```bash
tljh-config set limits.cpu None
```
and `config.py` will handle making it Python's `None`, and then your recommended change should work.

I'm sure there are other ways to handle this as well. Let me know what you think, @minrk.

jrdnbradford: @minrk looks like the `parse_value` method in `tljh/config.py` had logic to handle something like this but was a bit off. It checked `if value_str is None`, but there isn't a way to pass Python `None` directly to `tljh-config`. I updated that and the relevant tests.

jrdnbradford:
I went ahead and fixed the similar issue with the memory limit, which should allow `None`.

minrk:
Thanks!

PR #1015, opened 2024-12-09
Set development env vars in dev documentation correctly

yuvipanda:
Without this, the current dev setup documentation does not work.


minrk:
Nice!

PR #1022, opened 2025-02-14
Fix Ubuntu version in README

manics:
It's out of date, as reported in https://discourse.jupyter.org/t/installation-of-tljh-on-ubuntu-20-04-does-not-work/32823


yuvipanda:
ty @manics

PR #1024, opened 2025-03-07
Add `domains` `array` to `TLS` definition in config schema

jrdnbradford:
Fixes #1023.

```bash
sudo tljh-config add-item https.tls.domains sub-domain.xxx.yyyyyyy.edu
sudo tljh-config show
```
```yaml
user_environment:
  default_app: jupyterlab
https:
  enabled: true
  tls:
    domains:
    - sub-domain.xxx.yyyyyyy.edu
  ```

PR #1025, opened 2025-03-07
Fix `test_serverextensions` integration test

jrdnbradford:
Fix for integration test failures that seem to have started somewhat recently. See failures in #1024 and several recent `pre-commit autoupdate`s: https://github.com/jupyterhub/the-littlest-jupyterhub/actions/workflows/integration-test.yaml. 

Edit: see successful tests in my own fork: https://github.com/jrdnbradford/the-littlest-jupyterhub/actions/runs/13729823550

### DIFF
--- a/integration-tests/test_extensions.py
+++ b/integration-tests/test_extensions.py
@@ -9,11 +9,11 @@ def test_serverextensions():
     # jupyter-serverextension writes to stdout and stderr weirdly
     proc = subprocess.run(
         ["/opt/tljh/user/bin/jupyter-server", "extension", "list", "--sys-prefix"],
-        capture_output=True,
-        text=True,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT,
     )
 
-    output = proc.stdout + proc.stderr
+    output = proc.stdout.decode()
 
     extensions = [
         "jupyterlab",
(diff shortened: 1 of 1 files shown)


integration-tests/test_extensions.py
```diff
@@ -9,17 +9,20 @@ def test_serverextensions():
     # jupyter-serverextension writes to stdout and stderr weirdly
     proc = subprocess.run(
         ["/opt/tljh/user/bin/jupyter-server", "extension", "list", "--sys-prefix"],
-        stderr=subprocess.PIPE,
+        capture_output=True,
+        text=True,
     )
 
+    output = proc.stdout + proc.stderr
```
agoose77, line 16: Thanks for this change! It would be great to fix the integration tests.

One suggestion -- instead of using `capture_output` and concatenating the string objects, you could also set `stderr=subprocess.stdout` and check `assert 2 in proc.stdout.decode()`. 

In any case, this is mergeable, and I like the new assert string.

agoose77 (approved):

jrdnbradford:
Thanks, @agoose77! Much more elegant and less confusing, I think. This change has been pushed.

sgibson91:
Thanks @jrdnbradford for the work and @agoose77 for the review! 🚀

PR #1035, opened 2025-05-15
Fix default settings for culler instructions

paololazzari:
These are the settings in `values.yaml`:

```
cull:
  enabled: true
  users: false # --cull-users
  adminUsers: true # --cull-admin-users
  removeNamedServers: false # --remove-named-servers
  timeout: 3600 # --timeout
  every: 600 # --cull-every
  concurrency: 10 # --concurrency
  maxAge: 0 # --max-age
```


yuvipanda:
Thank you so much, @paololazzari!

PR #1069, opened 2026-04-29
Add Debian 13 support

jrdnbradford:
This PR: 

- adds tests for Debian 13 / Python 3.13
- updates relevant "supported OS" docs

Part of this setup uses code from the closed PR #1032 so that distros with `systemctl` do not halt the bootstrap script. In doing so it:

- fixes #1060
- fixes #1019
- seems relevant to #1066 where the issue was occurring on Debian 13 (although it is an oauthenticator issue)

I presume Debian 11 should remain tested until the LTS support ends 31 Aug 2026.

PR #1070, opened 2026-04-30
Consolidate supported operating systems in docs

jrdnbradford:
Based on a suggestion by @minrk in #1069.

Might be room for improvements here, but this is certainly better than the current documentation that can get stale easily.

The main OS requirements are described generally in `docs/topic/requirements.md`, with links to that heading where necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,8 +29,8 @@ more information.
 
 ## Installation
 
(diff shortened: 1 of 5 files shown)


minrk:
Thank you!

